PR #16, opened 2026-02-13
Claude image presets + env variable passing

jdc123:
Add --preset flag and "claude" preset for sandbox create
   Presets use embedded Dockerfiles (via Go's embed package) to build
   pre-configured container images. This approach is preferred over
   post-creation setup (docker exec) because Dockerfile layers are cached
   by Docker — the image only builds once, and subsequent sandbox
   creations are instant. It also ensures reproducible, declarative
   environments compared to imperative shell provisioning.

   The "claude" preset includes a full dev environment: Ubuntu 24.04,
   Node.js 22, Python 3, git, build-essential, and the Claude Code CLI.

   Also adds --env flag to sandbox create and materialize for passing
   environment variables (e.g. API keys) into containers.

   Usage:

     # First run builds the image (~1-2 min), then creates the container
     clawbox sandbox create --preset claude --name agent-sandbox

     # Subsequent runs use the cached image
     clawbox sandbox create --preset claude --name another-sandbox

     # Pass API key to the container
     clawbox sandbox create --preset claude --name my-sandbox \
       --env ANTHROPIC_API_KEY=...

### DIFF
--- a/cmd/clawbox/materialize.go
+++ b/cmd/clawbox/materialize.go
@@ -45,6 +45,7 @@ Examples:
 		destdir, _ := cmd.Flags().GetString("destdir")
 		image, _ := cmd.Flags().GetString("image")
 		mountStrs, _ := cmd.Flags().GetStringArray("mount")
+		envStrs, _ := cmd.Flags().GetStringArray("env")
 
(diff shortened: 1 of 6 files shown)


dbmikus (approved):
LGTM

PR #17, opened 2026-02-14
 Add interactive TTY mode and auto-mount Claude config in materialize

jdc123:
Add --interactive (-i) flag to materialize that allocates a pseudo-TTY
  and connects stdin, enabling interactive programs like Claude Code to
  run inside containers with full terminal UI support. In interactive mode,
  commands run directly (not wrapped in bash -c) so TTY signals propagate
  correctly, and --destdir becomes optional since the primary use case is
  the interactive session itself rather than file output.

  When --preset=claude is used, the host's Claude configuration is
  automatically propagated into the container by bind-mounting:
    - ~/.claude/  -> /root/.claude/  (rw)
    - ~/.claude.json -> /root/.claude.json (rw) This carries over auth tokens, settings, and project memory so Claude Code inside the container works with the user's existing session without manual setup. Both mounts are read-write so config changes made inside the container (e.g. new auth tokens) persist back to the host.

  Also updates tests to explicitly pass --image ubuntu:latest since the
  default image changed to clawbox-claude:latest, and removes trailing
  newline in the Claude preset Dockerfile.

### DIFF
--- a/cmd/clawbox/materialize.go
+++ b/cmd/clawbox/materialize.go
@@ -74,8 +74,8 @@ Examples:
 			}
 		}
 
-		if !interactive && destdir == "" {
-			return fmt.Errorf("--destdir is required (or use --interactive)")
+		if destdir == "" {
+			return fmt.Errorf("--destdir is required")
 		}
 
 		mounts, err := parseMountFlags(mountStrs)
--- a/cmd/clawbox/materialize.go
+++ b/cmd/clawbox/materialize.go
@@ -30,10 +30,13 @@ The --outdir flag controls which container directory gets copied to --destdir:
 Host directories can be mounted into the container using --mount:
   --mount /host/path:/container/path:ro
 
+Use --interactive (-i) to connect stdin/stdout for interactive programs like Claude:
+  clawbox materialize -i --cmd claude --mount $(pwd):/workspace
+
 Examples:
   clawbox materialize --cmd "echo hi > result.txt" --destdir /tmp/dest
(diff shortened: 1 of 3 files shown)


cmd/clawbox/materialize.go
```diff
@@ -44,21 +47,63 @@ Examples:
...
+					return err
+				}
+				fmt.Printf("Building %q preset image (this may take a few minutes)...\n", preset)
+				if err := sandbox.BuildDockerImage(image, dockerfile); err != nil {
+					return err
+				}
+			}
+		}
+
+		if !interactive && destdir == "" {
```
dbmikus: Shouldn't we also require a destination directory when running in interactive mode? If we aren't copying files out of the sandbox, what are we doing?

Technically, there is a reason to run an agent in a sandbox even if you don't copy files out, such as if you make MCP calls that modify external state, but don't think that's what we're going for here.

dbmikus (commented):
Even in interactive mode, I think we need to materialize files out of the sandbox back into the host in the destination directory. Otherwise I'm not sure what the point is of running the agent inside a sandboxed container, when we just remove the container at the end. Let me know if there was a good UX pattern here specifically that you were designing this for.

Also, it looks like you can't run a script when in interactive mode, but you can run a command. Is that correct? How does `-i` interplay with `--script` and/or `--cmd`. This makes sense to me.

cmd/clawbox/materialize.go
jdc123: Yeah good call

PR #48, opened 2026-03-08
Add per-repo .amika/config.toml support for sandbox setup script

dbmikus:
## Summary

- Introduce a new `internal/amikaconfig` package that loads per-repo configuration from `.amika/config.toml`
- When `--git` is used and `--setup-script` is not explicitly passed, automatically read `lifecycle.setup_script` from the repo's `.amika/config.toml` and bind-mount it into the container
- Document the `.amika/config.toml` schema and its interaction with `--setup-script` in `docs/sandbox-configuration.md`

## Motivation

Users want to commit sandbox configuration alongside their code so every collaborator gets a consistent container environment without having to pass extra CLI flags. By supporting a `.amika/config.toml` at the repo root, the setup script path becomes part of the repository itself rather than a per-user CLI invocation detail.

## Changes

- `internal/amikaconfig/config.go` — new package with `Config`, `LifecycleConfig` structs and `LoadConfig(repoRoot)` function using `github.com/BurntSushi/toml`
- `internal/amikaconfig/config_test.go` — unit tests covering missing file, valid config, malformed TOML, and empty lifecycle section
- `cmd/amika/sandbox.go` — integrate `amikaconfig.LoadConfig` into `sandbox create`; extract `setupScriptMountFromConfig` and `setupScriptBindMount` helpers to reduce duplication
- `docs/sandbox-configuration.md` — new "Per-repo configuration: `.amika/config.toml`" section with schema reference, path resolution rules, flag priority table, and example
- `go.mod` / `go.sum` — add `github.com/BurntSushi/toml v1.6.0` dependency

## Test plan

- [x] `make build` succeeds — both `dist/amika` and `dist/amika-server` produced
- [x] `go test ./internal/amikaconfig/...` — all four unit tests pass (missing file, valid config, malformed TOML, empty lifecycle section)
- [x] All other package tests pass; the one pre-existing failure in `internal/app` (`TestNewService_ImplementsPublicService`) is unrelated to this PR — it is caused by a malformed local `~/.claude/.credentials.json` and reproduces identically on `main`
- [x] `--git` without `--setup-script` uses `.amika/config.toml`: verified by code — `gitMountInfo != nil && !cmd.Flags().Changed("setup-script")` guard at line 107 of `sandbox.go`
- [x] `--setup-script` overrides `.amika/config.toml`: verified by code — when `--setup-script` is set `cmd.Flags().Changed("setup-script")` is true, so the config block is skipped entirely and the flag mount is appended instead
- [x] `amika sandbox create` without `--git` ignores `.amika/config.toml`: verified by code — `gitMountInfo` is nil when `--git` is not passed, so the config block is never reached

### DIFF
--- a/cmd/amika/sandbox.go
+++ b/cmd/amika/sandbox.go
@@ -105,19 +105,12 @@ var sandboxCreateCmd = &cobra.Command{
 		}
 		// Read .amika/config.toml from git repo root, unless --setup-script was explicitly provided.
 		if gitMountInfo != nil && !cmd.Flags().Changed("setup-script") {
-			cfg, err := amikaconfig.LoadConfig(gitMountInfo.RepoRoot)
+			mount, err := setupScriptMountFromConfig(gitMountInfo.RepoRoot)
 			if err != nil {
-				return fmt.Errorf("failed to read .amika/config.toml: %w", err)
+				return err
 			}
-			if cfg != nil && cfg.Lifecycle.SetupScript != "" {
-				scriptPath := cfg.Lifecycle.SetupScript
-				if !filepath.IsAbs(scriptPath) {
-					scriptPath = filepath.Join(gitMountInfo.RepoRoot, scriptPath)
-				}
-				if _, err := os.Stat(scriptPath); err != nil {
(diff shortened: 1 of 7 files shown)


jdc123 (approved):
lgtm

PR #51, opened 2026-03-09
Fix preset image auto-build and fix ImageFlagChanged 

jdc123:
Fix preset image auto-build for amika-server sandbox creation

The public service's CreateSandbox (pkg/amika) passed the image name
directly to Docker without calling ResolveAndEnsureImage, so preset
images (e.g. amika/coder) were never auto-built when using amika-server.
This caused a 503 "Unable to find image" error. The CLI already handled
this correctly; this brings the server path in line.

Fix ImageFlagChanged to not override preset when image is also set

When both Image and Preset are provided in CreateSandboxRequest,
ImageFlagChanged was incorrectly set to true, causing the preset
to be ignored. Now ImageFlagChanged is only true when an explicit
image is given without a preset.

### DIFF
--- a/pkg/amika/service.go
+++ b/pkg/amika/service.go
@@ -85,6 +85,17 @@ func (s *serviceImpl) CreateSandbox(_ context.Context, req CreateSandboxRequest)
 		return Sandbox{}, err
 	}
 
+	resolvedImage, err := sandbox.ResolveAndEnsureImage(sandbox.PresetImageOptions{
+		Image:              req.Image,
+		Preset:             req.Preset,
+		ImageFlagChanged:   req.Image != "",
+		DefaultBuildPreset: "coder",
+	})
+	if err != nil {
+		return Sandbox{}, fmt.Errorf("%w: %v", ErrDependency, err)
+	}
+	req.Image = resolvedImage.Image
+
 	mounts := toSandboxMountBindings(req.Mounts, req.Volumes)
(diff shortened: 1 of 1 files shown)


dbmikus:
I think this is the wrong logic?

> When both Image and Preset are provided in CreateSandboxRequest, ImageFlagChanged was incorrectly set to true, causing the preset to be ignored. Now ImageFlagChanged is only true when an explicit image is given without a preset.

If the user sets the `Image` flag, the `ImageFlagChanged` should be true. At that point, we should use the provided Docker image. I can't remember if there is anything more we want to preserve from the preset, but IMO an image flag should override a preset flag, as the image is more specific.

I'll look at the code.

pkg/amika/service.go
```diff
@@ -85,6 +85,17 @@ func (s *serviceImpl) CreateSandbox(_ context.Context, req CreateSandboxRequest)
 		return Sandbox{}, err
 	}
 
+	resolvedImage, err := sandbox.ResolveAndEnsureImage(sandbox.PresetImageOptions{
+		Image:              req.Image,
+		Preset:             req.Preset,
+		ImageFlagChanged:   req.Image != "" && req.Preset == "",
```
dbmikus: Feels like `ImageFlagChanged` should jsut be whether `req.Image` was set. Am I missing something?

PR #53, opened 2026-03-10
Add CLI authentication via WorkOS Device Authorization Flow

jdc123:
Implement `amika auth login`, `amika auth logout`, and `amika auth status` commands using the OAuth 2.0 Device Authorization Flow (RFC 8628).

- login: opens browser for device authorization, polls for token, persists session to ~/.local/state/amika/workos-session.json
- logout: deletes the persisted session file
- status: displays current user email and org, or "Not logged in"
- Auto-refreshes expired access tokens using the stored refresh token

### DIFF
--- a/internal/auth/device_flow.go
+++ b/internal/auth/device_flow.go
@@ -1,0 +1,184 @@
+package auth
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os/exec"
+	"runtime"
+	"strings"
+	"time"
+)
+
+const workosDeviceAuthorizeURL = "https://api.workos.com/user_management/authorize/device"
+
+type deviceCodeResponse struct {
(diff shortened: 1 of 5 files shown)


internal/auth/device_flow.go
```diff
@@ -0,0 +1,184 @@
...
+
+		var errResp struct {
+			Error string `json:"error"`
+		}
+		if err := json.Unmarshal(respBody, &errResp); err != nil {
+			continue
+		}
+
+		switch errResp.Error {
+		case "authorization_pending":
```
dbmikus, line 120: Are these the real error responses?

dbmikus (commented):
Overall, LGTM! Had a question about error types.

internal/auth/device_flow.go
jdc123: Yup, https://workos.com/docs/authkit/cli-auth#3-request-tokens

PR #56, opened 2026-03-11
Add remote sandbox support with local/remote mode routing

jdc123:
- Add API client (internal/apiclient) for the remote Amika API (POST/GET/DELETE /api/sandboxes) with bearer token auth
- Add --local and --remote persistent flags to sandbox commands
- When logged in, sandbox list merges local and remote results; sandbox create/delete route to remote by default
- When not logged in, commands operate locally with a notice
- Add Location column to sandbox list output (local/remote)
- Add --github-url flag for remote sandbox creation; auto-derives GitHub URL from git origin when --git is used with remote mode
- Remote API URL defaults to https://api.amika.dev, overridable via AMIKA_API_URL env var

### DIFF
--- a/cmd/amika/sandbox.go
+++ b/cmd/amika/sandbox.go
@@ -334,17 +334,27 @@ var sandboxDeleteCmd = &cobra.Command{
 
 		var errs []string
 		for _, name := range args {
-			info, localErr := store.Get(name)
-			if localErr != nil && (mode == "remote" || mode == "both") {
-				// Not found locally, try remote.
+			// Remote-only mode: skip local entirely.
+			if mode == "remote" {
 				if remoteClient != nil {
 					if remoteErr := remoteClient.DeleteSandbox(name); remoteErr != nil {
 						errs = append(errs, fmt.Sprintf("sandbox %q: %v", name, remoteErr))
-						continue
+					} else {
+						fmt.Printf("Sandbox %q deleted (remote)\n", name)
 					}
-					fmt.Printf("Sandbox %q deleted (remote)\n", name)
-					continue
 				}
+				continue
+			}
+
+			info, localErr := store.Get(name)
+			if localErr != nil && mode == "both" && remoteClient != nil {
+				// Not found locally, try remote.
+				if remoteErr := remoteClient.DeleteSandbox(name); remoteErr != nil {
+					errs = append(errs, fmt.Sprintf("sandbox %q: %v", name, remoteErr))
+				} else {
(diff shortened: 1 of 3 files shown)


cmd/amika/sandbox.go
```diff
@@ -258,10 +320,30 @@ var sandboxDeleteCmd = &cobra.Command{
...
+
 		var errs []string
 		for _, name := range args {
-			info, err := store.Get(name)
-			if err != nil {
+			info, localErr := store.Get(name)
+			if localErr != nil && (mode == "remote" || mode == "both") {
+				// Not found locally, try remote.
+				if remoteClient != nil {
+					if remoteErr := remoteClient.DeleteSandbox(name); remoteErr != nil {
```
dbmikus, line 340: If `mode` is "remote" or "both", but the sandbox in question is "local", is this going to cause an error?

cmd/amika/sandbox.go
```diff
@@ -1297,8 +1473,8 @@ func init() {
 	sandboxCreateCmd.Flags().Bool("yes", false, "Skip mount confirmation prompt")
 	sandboxCreateCmd.Flags().Bool("connect", false, "Connect to the sandbox shell immediately after creation")
 	sandboxCreateCmd.Flags().String("setup-script", "", "Mount a local script file to /opt/setup.sh in the container (read-only)")
+	sandboxCreateCmd.Flags().String("github-url", "", "GitHub repository URL for remote sandbox creation")
```
dbmikus: I think we can delete this.

cmd/amika/sandbox.go
```diff
@@ -1274,13 +1386,77 @@ func resolveDeleteVolumes(
 	return confirmed, nil
 }
 
+func createRemoteSandbox(cmd *cobra.Command) error {
+	name, _ := cmd.Flags().GetString("name")
+	githubURL, _ := cmd.Flags().GetString("github-url")
```
dbmikus: I think we can delete this argument. The user should be able to specify a git url via `--git <url>`, or they should be able to specify a local repo. If we fail to extract the remote git URL when acting locally, we can print an error message to the user telling them to specify a "git http(s) or ssh URL"

dbmikus:
My feedback here is that we should not add a `--github-url` and can get the `--git` argument to behave how we want.

cmd/amika/sandbox.go
jdc123: Good catch, fixing

dbmikus (approved):
LGTM

PR #62, opened 2026-03-12
Add sandbox ssh command for remote SSH access

jdc123:
- Add `amika sandbox ssh <name>` to connect to remote sandboxes via SSH
- Fetches SSH destination from POST /api/sandboxes/{name}/ssh and execs into an interactive ssh session
- Add --revoke flag to revoke SSH access (DELETE /api/sandboxes/{name}/ssh)
- Local sandboxes get a clear error: "SSH access currently only works for remote sandboxes"
- Add GetSSH and RevokeSSH methods to the API client

### DIFF
--- a/cmd/amika/sandbox.go
+++ b/cmd/amika/sandbox.go
@@ -1471,12 +1471,85 @@ func resolveGitURL(value string) (string, error) {
 	return origin, nil
 }
 
+var sandboxSSHCmd = &cobra.Command{
+	Use:   "ssh [name]",
+	Short: "SSH into a remote sandbox",
+	Long: `Connect to a remote sandbox via SSH, or revoke SSH access.
+
+Examples:
+  amika sandbox ssh my-sandbox
+  amika sandbox ssh --name my-sandbox
+  amika sandbox ssh my-sandbox --revoke`,
+	Args: cobra.MaximumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+		cmd.SilenceErrors = true
+
+		name, _ := cmd.Flags().GetString("name")
+		if len(args) > 0 {
+			name = args[0]
+		}
+		if name == "" {
+			return fmt.Errorf("sandbox name is required")
+		}
+
+		// Check if this is a local sandbox.
+		sandboxesFile, err := config.SandboxesStateFile()
+		if err == nil {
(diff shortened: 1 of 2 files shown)


cmd/amika/sandbox.go
```diff
@@ -1471,12 +1471,85 @@ func resolveGitURL(value string) (string, error) {
...
 }
 
+var sandboxSSHCmd = &cobra.Command{
+	Use:   "ssh [name]",
+	Short: "SSH into a remote sandbox",
+	Long: `Connect to a remote sandbox via SSH, or revoke SSH access.
+
+Examples:
+  amika sandbox ssh my-sandbox
+  amika sandbox ssh --name my-sandbox
```
dbmikus: Seems a little confusing that there's two ways to specify the name. I'd just do `amika sandbox ssh <name>`

cmd/amika/sandbox.go
```diff
@@ -1471,12 +1471,85 @@ func resolveGitURL(value string) (string, error) {
...
+  amika sandbox ssh my-sandbox --revoke`,
+	Args: cobra.MaximumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+		cmd.SilenceErrors = true
+
+		name, _ := cmd.Flags().GetString("name")
+		if len(args) > 0 {
+			name = args[0]
+		}
```
dbmikus: If we did decide to keep both `amika sandbox ssh {name}` and `amika sandbox ssh {name} --revoke`, then I would expect it to be an error to specify both a positional name and a CLI flag name.

That said, I advocate for getting rid of the `--name` flag.

cmd/amika/sandbox.go
```diff
@@ -1501,4 +1574,6 @@ func init() {
 	sandboxDeleteCmd.Flags().Bool("delete-volumes", false, "Also delete associated volumes that are no longer referenced")
 	sandboxDeleteCmd.Flags().Bool("keep-volumes", false, "Keep associated volumes even when only this sandbox references them")
 	sandboxConnectCmd.Flags().String("shell", "zsh", "Shell to run in the sandbox container")
+	sandboxSSHCmd.Flags().String("name", "", "Name of the sandbox to SSH into")
```
dbmikus: Would get rid of the `--name` flag, IMO

dbmikus (approved):
LGTM

PR #79, opened 2026-03-16
feat: add secrets extract and push commands for managing remote secrets

jdc123:
Adds `amika secrets` subcommands for discovering local credentials and pushing secrets to the remote Amika secrets store.

- `amika secrets extract` discovers local API credentials and displays them with masked values. `--push` pushes after interactive y/N confirmation. `--only` filters which secrets are included.
- `amika secrets push KEY=VALUE` pushes explicitly provided secrets. `--from-env=KEY1,KEY2` reads values from the current shell environment.
- Both commands create or update secrets by name via /api/secrets endpoints.
- Removes `auth extract` subcommand (replaced by `secrets extract`).
- Adds ListSecrets, CreateSecret, UpdateSecret methods to the API client.


cmd/amika/secrets_test.go
dbmikus, line 1: nit, but we should move all the `amika secrets extract` tests to `cmd/amika/secrets_test.go`.

cmd/amika/secrets.go
```diff
@@ -0,0 +1,314 @@
...
+			if remote, ok := existingByName[key]; ok {
+				err = client.UpdateSecret(remote.ID, apiclient.UpdateSecretRequest{
+					Value: value,
+				})
+				if err != nil {
+					return fmt.Errorf("updating secret %s: %w", key, err)
+				}
+				fmt.Fprintf(cmd.OutOrStdout(), "  Updated %s\n", key)
+			} else {
+				err = client.CreateSecret(apiclient.CreateSecretRequest{
```
dbmikus: We should default to "user" scope unless they specify "--scope org" or something like that.

Even if the API Server implicitly uses "user" scope, we should be explicit about it here.

cmd/amika/secrets.go
```diff
@@ -0,0 +1,314 @@
...
+			existingByName[s.Name] = s
+		}
+
+		// Push each secret.
+		for _, key := range keys {
+			value := secrets[key]
+			if remote, ok := existingByName[key]; ok {
+				err = client.UpdateSecret(remote.ID, apiclient.UpdateSecretRequest{
+					Value: value,
+				})
```
dbmikus: If secrets already exist, we should check if the scope of the existing secret matches the scope we are trying to push (defaulting to "user"). If they don't match, then I think we should return an error, printing about scope mismatch and ask the user to specify a "--scope (user|org)"

cmd/amika/secrets.go
```diff
@@ -0,0 +1,314 @@
...
+				if err != nil {
+					return fmt.Errorf("updating secret %s: %w", key, err)
+				}
+				fmt.Fprintf(cmd.OutOrStdout(), "  Updated %s\n", key)
+			} else {
+				err = client.CreateSecret(apiclient.CreateSecretRequest{
```
dbmikus: I think we should explicitly put a default user scope here.

dbmikus (commented):
Can you also tell the AI agent to update the `docs/`?

cmd/amika/secrets.go
jdc123: Great point.

cmd/amika/secrets.go
jdc123: Good point

dbmikus (approved):
LGTM

PR #88, opened 2026-03-18
WIP Add session TTL support and custom duration parsing

dbmikus:
## Summary
This PR adds support for configurable session time-to-live (TTL) in the authentication flow, allowing users to specify how long a login session remains valid before requiring re-authentication. It also introduces a human-friendly duration parser to support flexible time specifications.

## Key Changes

- **New duration parser** (`internal/auth/duration.go`): Added `ParseDuration()` function that parses human-friendly duration strings with support for seconds (s), minutes (min), hours (h), days (d), and weeks (w). Includes comprehensive test coverage for valid inputs and error cases.

- **Session TTL tracking**: Extended `WorkOSSession` struct with a new `SessionExpiresAt` field to track when the overall session expires, separate from the access token expiration.

- **Device login enhancement**: Modified `DeviceLogin()` to accept a `sessionTTL` parameter, with a default of 7 days (`DefaultSessionTTL`). The TTL is propagated through the token polling and parsing flow.

- **Refresh token handling**: Updated `RefreshAccessToken()` to accept the full session object instead of just the refresh token, allowing it to preserve the original `SessionExpiresAt` across token refreshes.

- **Session validation**: Enhanced `GetValidSession()` to check if the overall session has expired (not just the access token), returning an error if re-login is required.

- **CLI integration**: Added `--ttl` flag to the `auth login` command, allowing users to specify custom session lifetimes (e.g., `--ttl 2h`, `--ttl 30d`).

## Implementation Details

- The duration parser uses longest-suffix matching to correctly handle "min" before "m" to avoid ambiguity.
- Negative and zero durations are rejected as invalid.
- Session expiration is checked before attempting token refresh, providing clearer error messaging.
- The `SessionExpiresAt` field is preserved across token refreshes to maintain the original session lifetime.

https://claude.ai/code/session_01RWDuSnj9JVUg867CKjSbZH

### DIFF
--- a/internal/auth/device_flow.go
+++ b/internal/auth/device_flow.go
@@ -24,7 +24,12 @@ type deviceCodeResponse struct {
 }
 
 // DeviceLogin performs the OAuth 2.0 Device Authorization Flow via WorkOS.
-func DeviceLogin(clientID string) (*WorkOSSession, error) {
+// The sessionTTL controls how long the overall session remains valid before
+// the user must log in again. Pass 0 to use DefaultSessionTTL.
+func DeviceLogin(clientID string, sessionTTL time.Duration) (*WorkOSSession, error) {
+	if sessionTTL == 0 {
+		sessionTTL = DefaultSessionTTL
+	}
 	// Step 1: Request device code.
 	dc, err := requestDeviceCode(clientID)
 	if err != nil {
@@ -38,7 +43,7 @@ func DeviceLogin(clientID string) (*WorkOSSession, error) {
 	openBrowser(dc.VerificationURIComplete)
 
 	// Step 3: Poll for token.
-	return pollForToken(clientID, dc)
+	return pollForToken(clientID, dc, sessionTTL)
 }
 
 func requestDeviceCode(clientID string) (*deviceCodeResponse, error) {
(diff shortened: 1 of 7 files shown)


internal/auth/device_flow.go
```diff
@@ -154,12 +159,13 @@ func parseTokenResponse(respBody []byte) (*WorkOSSession, error) {
...
-		Email:        result.User.Email,
-		OrgID:        result.OrganizationID,
-		ExpiresAt:    expiresAt,
+		AccessToken:      result.AccessToken,
+		RefreshToken:     result.RefreshToken,
+		UserID:           result.User.ID,
+		Email:            result.User.Email,
+		OrgID:            result.OrganizationID,
+		ExpiresAt:        expiresAt,
+		SessionExpiresAt: time.Now().Add(sessionTTL),
```
dbmikus, line 168: In our struct type, can you document the difference between these?

dbmikus:
I'm not sure that this is the right approach. We should probably just make something that handles auth refresh if we get rejected to due an expired token.

dbmikus:
Not 100% confirmed, but I don't think we can control this. We can set oauth lifetime from WorkOS.

PR #89, opened 2026-03-18
fix: auto-refresh WorkOS auth and centralize env var resolution

dbmikus:
## Summary
- Add `TokenSource` interface to `apiclient` so the API client auto-refreshes WorkOS sessions before each request, instead of callers manually fetching tokens upfront
- Move WorkOS session file path resolution into the `basedir.Paths` interface, following the existing XDG pattern
- Centralize `AMIKA_WORKOS_CLIENT_ID` and `AMIKA_API_URL` env var resolution into `internal/config`, fixing a bug where `auth login` honored the env var but all other callers (sandbox, secrets) used the hardcoded default — causing token refresh to fail when the env var was set

### DIFF
--- a/cmd/amika/auth.go
+++ b/cmd/amika/auth.go
@@ -2,13 +2,11 @@ package main
 
 import (
 	"fmt"
-	"os"
 
 	"github.com/gofixpoint/amika/internal/auth"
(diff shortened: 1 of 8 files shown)


jdc123 (approved):
lgtm

PR #92, opened 2026-03-20
feat: add `sandbox code` command to open remote sandboxes in Cursor

jdc123:
(via ssh)

Add a new `amika sandbox code <name>` subcommand that opens a remote sandbox in Cursor using its SSH remote extension. The command reuses the same SSH destination logic from `sandbox ssh`, extracts the user@host portion, and invokes `cursor --remote ssh-remote+<host>`.

Includes error handling for missing Cursor CLI, unsupported editors, and missing Remote SSH extension. Defaults to Cursor via the optional `--editor` flag.

### DIFF
--- a/cmd/amika/sandbox.go
+++ b/cmd/amika/sandbox.go
@@ -1672,12 +1672,18 @@ Examples:
 		// The last field is the user@host (or host) destination.
 		remoteHost := fields[len(fields)-1]
 
-		cursorCmd := exec.Command("cursor", "--remote", "ssh-remote+"+remoteHost, "/home/amika/workspace")
+		remotePath := "/home/amika/workspace"
+		if info.RepoName != "" {
+			remotePath = remotePath + "/" + info.RepoName
+		}
+
+		cursorCmd := exec.Command("cursor", "--remote", "ssh-remote+"+remoteHost, remotePath)
 		cursorCmd.Stdin = os.Stdin
 		cursorCmd.Stdout = os.Stdout
 		cursorCmd.Stderr = os.Stderr
 
 		fmt.Fprintf(cmd.OutOrStdout(), "Opening sandbox %q in Cursor via SSH (%s)...\n", name, remoteHost)
+		fmt.Fprintf(cmd.OutOrStdout(), "Running: cursor --remote ssh-remote+%s %s\n", remoteHost, remotePath)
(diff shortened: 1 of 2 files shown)


cmd/amika/sandbox.go
```diff
@@ -1610,6 +1610,83 @@ Examples:
...
+
+		// Build the ssh-remote+ URI for cursor.
+		// SSHDestination may include flags (e.g. "-o StrictHostKeyChecking=no user@host"),
+		// but cursor needs just the host/user@host portion as the remote identifier.
+		sshDest := info.SSHDestination
+		fields := strings.Fields(sshDest)
+		// The last field is the user@host (or host) destination.
+		remoteHost := fields[len(fields)-1]
+
+		cursorCmd := exec.Command("cursor", "--remote", "ssh-remote+"+remoteHost, "/home/amika/workspace")
```
jdc123: We may want to override the remote path (`/home/amika/workspace`) in the future

dbmikus: Two things:

1. ideally, we connect to the repo directory precisely (`/home/amika/workspace/${repo}`)
2. right now, IMO we should connect to `/home/amika` because that's where `amika sandbox ssh` connects to, and it's the default behavior for connecting to remote machines

Currently, we have a behavior to dynamically set the agent's current working directory, which we use for configuring OpenCode. It is set to the repo path that we check out on the sandbox. I think we could try to load that data when running the `amika code` command, so that it connects to the right place.

Some options for doing that:

1. store the value in the `app.sandboxes` table
2. save the `AMIKA_AGENT_CWD` environment variable on the sandbox and check that value. You probably want to set it somewhere in the filesystem

(1) seems like the much simpler solution. (2) is a bit more decentralized and "git-like", aka federated, where there doesn't need to be a central control plane. I would not do (2) right now, IMO. I think (2) becomes viable when we have `amikad`.

If you do opt for (2), read up on the [Filesystem Hierarchy Standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) (more [detailed guide](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html) from Linux Foundation), and optionally the [XDG Base Directory](https://wiki.archlinux.org/title/XDG_Base_Directory) spec.

dbmikus (approved):
LGTM

dbmikus (approved):

PR #102, opened 2026-03-23
feat: add `--secret` flag to `sandbox create` for remote secret injection

dbmikus:
Parse `--secret env:FOO=SECRET_NAME` (or shorthand `env:SECRET_NAME`) flags and pass them through to the remote API when creating sandboxes. Also wire `--env` flags into the remote create request as `env_vars`.

Using `--secret` in local mode returns a clear error since secrets are resolved server-side.


jdc123 (approved):
lgtm

PR #104, opened 2026-03-23
WIP Add a spec for fcp command to copy files

dbmikus:


specs/006-sandbox-file-copy.md
```diff
@@ -0,0 +1,91 @@
...
+### Arguments
+
+Exactly two positional arguments: `<from>` and `<to>`. Exactly one of them must reference a sandbox using the format `<sandbox-name>:<path>`. The other is a local filesystem path.
+
+| Argument     | Format              | Example                           |
+| ------------ | ------------------- | --------------------------------- |
+| Local path   | Any filesystem path | `./file.txt`, `/tmp/output/`      |
+| Sandbox path | `<name>:<path>`     | `my-sandbox:/home/amika/file.txt` |
+
+**Parsing heuristic:** A colon-separated argument is treated as a sandbox reference only if the portion before the first `:` contains no `/` character. This correctly distinguishes sandbox references from absolute or relative local paths (e.g. `/foo/bar`, `./foo`).
```
dbmikus, line 27: This is wrong. The syntax is:

```
amika sandbox fcp [flags] <from> <to>
```

Each of `<from>` and `<to>` are "targets". A target can be of one of two format:

1. `path`: a normal filepath like `/foo/bar/baz`
2. `sbox:path`: `path` is same as above, `sbox` is a sandbox name. They are separated by a colon. We have separate rules for what values can be in a sandbox name. Notably, the ":" and "@" characters are not allowed, but characters like from this regex `[a-zA-Z0-9./_-]` are allowed.

PR #106, opened 2026-03-23
feat: add `--from-file` flag to `secret push` for .env file support

dbmikus:
Allow `amika secret push --from-file=.env` to load secrets from a .env file using Docker-style parsing (skip comments and blank lines, split on first `=`, values taken verbatim). When combined with other sources, --from-file is lowest priority, then positional args, then --from-env.


jdc123 (approved):
lgtm

PR #107, opened 2026-03-23
feat: make `sandbox create --connect` work for remote sandboxes

jdc123:
The --connect flag was ignored in the remote sandbox creation path because createRemoteSandbox() returned before the flag was checked.

Extract shared execSSH() helper from duplicated SSH logic in sandbox ssh, sandbox connect, and sandbox create --connect.


dbmikus (approved):
LGTM

PR #110, opened 2026-03-24
feat: push sized Daytona snapshots (xs, m) with explicit resource specs

dbmikus:
Push each Daytona image as multiple snapshots with different resource allocations instead of a single default snapshot. Sizes: xs (1 vCPU, 1GB RAM, 3GB disk) and m (2 vCPU, 8GB RAM, 10GB disk). Snapshot names include the size suffix, e.g. amika/daytona-coder-xs:{version}.


jdc123 (approved):
lgtm

PR #111, opened 2026-03-24
feat: update sandbox create API to use secret_env_vars, preset, and size

jdc123:
The remote API replaced the secrets array with a secret_env_vars map (env var name → secret ID) and added preset/size fields. This updates the CLI to match:

- Replace Secrets []SecretRef with SecretEnvVars map[string]string
- Resolve --secret flag names to IDs via ListSecrets before sending
- Pass --preset and --size flags through to remote sandbox creation
- Auto-generate sandbox name when --name is omitted (remote mode)

### DIFF
--- a/cmd/amika/sandbox.go
+++ b/cmd/amika/sandbox.go
@@ -849,18 +849,15 @@ func parseVolumeFlags(flags []string) ([]sandbox.MountBinding, error) {
 	return mounts, nil
 }
 
-// secretMapping holds a parsed --secret flag entry before resolution.
-type secretMapping struct {
-	envVar     string
-	secretName string
-}
-
-// parseSecretFlags parses --secret flag values. Supported syntax:
+// parseSecretFlags parses --secret flag values into a map of env var name → secret name.
+// Supported syntax:
 //   - env:FOO=SECRET_NAME — inject secret SECRET_NAME as env var FOO
 //   - env:SECRET_NAME     — shorthand: env var name equals the secret name
-func parseSecretFlags(flags []string) ([]secretMapping, error) {
-	var mappings []secretMapping
-	seenEnv := make(map[string]bool)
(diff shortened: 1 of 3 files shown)


cmd/amika/sandbox.go
```diff
@@ -894,12 +900,39 @@ func parseSecretFlags(flags []string) ([]apiclient.SecretRef, error) {
...
+			secretName: secretName,
 		})
 	}
-	return refs, nil
+	return mappings, nil
+}
+
+// resolveSecretEnvVars resolves secret names to IDs and returns a map of env var name → secret ID
+// suitable for the secret_env_vars field in the API request.
+func resolveSecretEnvVars(client *apiclient.Client, mappings []secretMapping) (map[string]string, error) {
```
dbmikus: Ideally, we'd be able to just send the secret names to the API Server and it would handle mapping them to IDs. I thought that's how the API Server worker? What is the "value" part of the [`secret_env_vars`](https://github.com/gofixpoint/amika-mono/blob/2609cba0cb416f8265dccf147d4d48a130d2178c/js/coding-agents/src/lib/schemas/sandbox.ts#L33) in a `CreateSandboxRequest`? Is it a secret ID or a secret name?

This is what I was asking about in https://github.com/gofixpoint/amika-mono/pull/162#discussion_r2982879229

I thought you said the API Server accepted secret names, instead of IDs, but I must have misunderstood. To make sure I understand, is it the case that in the statement `CreateSandboxRequest.secret_env_vars[key] = value`, the `value` should be a:

1. secret name
2. secret ID

Whether it's (1) or (2), I don't care right now as long as the user interface works with secret names. But I'd like to understand.

dbmikus (approved):
LGTM but I have an inline question

dbmikus (approved):
LGTM

PR #112, opened 2026-03-25
feat: send --setup-script contents as setup_script_text for remote sandboxes

dbmikus:
The --setup-script flag was silently ignored for remote sandbox creation. Now reads the local file and sends its contents via the setup_script_text API field.


jdc123 (approved):
lgtm

PR #116, opened 2026-03-25
feat: add .amika/config.toml with Go install setup script

dbmikus:
## Summary
- Adds `.amika/config.toml` with a lifecycle setup script (no services)
- Adds `.amika/scripts/setup.sh` that checks if Go is installed and installs Go 1.26.1 if missing
- Supports both amd64 and arm64 architectures

## Test plan
- [ ] Verify `amika sandbox create --git` picks up the config and mounts the setup script
- [ ] Test in a sandbox without Go pre-installed to confirm installation works
- [ ] Test in a sandbox with Go already installed to confirm it skips installation

### DIFF
--- a/.amika/config.toml
+++ b/.amika/config.toml
@@ -1,0 +1,2 @@
+[lifecycle]
+setup_script = ".amika/scripts/setup.sh"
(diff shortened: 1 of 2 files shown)


jdc123 (approved):
lgtm

PR #118, opened 2026-03-26
feat: poll sandbox create endpoint until provisioning completes

jdc123:
The remote sandbox create API is now asynchronous (returns 202 with state "initializing"). Add GetSandbox and WaitForSandbox to the API client, and update the CLI to poll every 3s until the sandbox reaches a ready state.


dbmikus (approved):
LGTM

PR #119, opened 2026-03-27
feat: add `secret claude` commands for credential upload, list, and delete

jdc123:
Add CLI commands to manage Claude Code credentials for sandbox authentication:

- `secret claude upload`: Interactive discovery of local credentials (API keys from config files, OAuth from credential files and macOS Keychain). Users see what's available, pick one, and confirm before uploading. Supports --type api_key (reads ANTHROPIC_API_KEY env var) and --type oauth (reads keychain on macOS, credential files on Linux). Also accepts --value and --from-file for direct input.

- `secret claude list`: Lists uploaded credentials with ID, name, and type.

- `secret claude delete <id>`: Deletes a credential by ID.

Sandbox creation now auto-selects the Claude credential by name when exactly one exists, passing claude_credential_name in the create request.

API client additions:
- POST /api/secrets/claude (name required, unique per user, type: oauth|api_key)
- GET /api/secrets/claude
- DELETE /api/secrets/:id
- ClaudeCredentialName field on CreateSandboxRequest

Auth package additions:
- DiscoverClaudeCredentials() returns all local Claude credentials with their type and source preserved for interactive selection.

### DIFF
--- a/cmd/amika/sandbox.go
+++ b/cmd/amika/sandbox.go
@@ -1811,15 +1811,19 @@ func createRemoteSandbox(cmd *cobra.Command, target string) error {
 		setupScriptText = string(data)
 	}
 
+	// Auto-select Claude credential if one exists.
+	claudeCredentialName := autoSelectClaudeCredential(client)
+
 	req := apiclient.CreateSandboxRequest{
-		Name:            name,
-		Provider:        "daytona",
-		GitHubURL:       gitURL,
-		EnvVars:         envVars,
-		SecretEnvVars:   secretEnvVars,
-		Preset:          preset,
-		Size:            size,
-		SetupScriptText: setupScriptText,
+		Name:                 name,
+		Provider:             "daytona",
+		GitHubURL:            gitURL,
+		EnvVars:              envVars,
+		SecretEnvVars:        secretEnvVars,
+		Preset:               preset,
+		Size:                 size,
+		SetupScriptText:      setupScriptText,
+		ClaudeCredentialName: claudeCredentialName,
 	}
 
 	sb, err := client.CreateSandbox(req)
@@ -1848,6 +1852,17 @@ func createRemoteSandbox(cmd *cobra.Command, target string) error {
 // remote sandbox creation. If the value is already an HTTP(S) or SSH URL, it is
 // returned directly. Otherwise it is treated as a local path and the origin
 // remote URL is extracted.
+// autoSelectClaudeCredential returns the name of the user's Claude credential
+// if exactly one exists on the remote. Returns empty string on any error or
+// if no credentials are uploaded.
+func autoSelectClaudeCredential(client *apiclient.Client) string {
+	creds, err := client.ListClaudeSecrets()
+	if err != nil || len(creds) != 1 {
+		return ""
+	}
+	return creds[0].Name
+}
+
 func resolveGitURL(value string) (string, error) {
 	// Already a URL — use as-is.
 	if strings.HasPrefix(value, "http://") || strings.HasPrefix(value, "https://") || strings.HasPrefix(value, "git@") {
--- a/cmd/amika/secrets.go
+++ b/cmd/amika/secrets.go
@@ -2,8 +2,13 @@ package main
 
 import (
 	"bufio"
+	"encoding/json"
 	"fmt"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strconv"
 	"strings"
 	"text/tabwriter"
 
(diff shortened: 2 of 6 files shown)


cmd/amika/sandbox.go
```diff
@@ -1848,6 +1852,17 @@ func createRemoteSandbox(cmd *cobra.Command, target string) error {
 // remote sandbox creation. If the value is already an HTTP(S) or SSH URL, it is
 // returned directly. Otherwise it is treated as a local path and the origin
 // remote URL is extracted.
+// autoSelectClaudeCredential returns the name of the user's Claude credential
+// if exactly one exists on the remote. Returns empty string on any error or
+// if no credentials are uploaded.
```
dbmikus, line 1857: Comments are messed up. You need to put the `// resolveGitURL` comments above the `resolveGitURL` function

cmd/amika/secrets.go
```diff
@@ -398,12 +403,340 @@ func maskValue(value string) string {
...
+
+			nameFlag, _ := cmd.Flags().GetString("name")
+			value, _ := cmd.Flags().GetString("value")
+			fromFile, _ := cmd.Flags().GetString("from-file")
+			typeFlag, _ := cmd.Flags().GetString("type")
+
+			var credValue string
+			var credType string // "oauth" or "api_key"
+
+			switch {
```
dbmikus, line 440: If it's mutually exclusive to set `--value`, `--from-file`, or `--type`, then I think we should explicitly error out if the user sets multiple, instead of implicitly having some order in which we prefer those flags.

dbmikus: Also, put this inside a function call `parseClaudeCreds` that returns `credValue, credType, err`

cmd/amika/secrets.go
```diff
@@ -398,12 +403,340 @@ func maskValue(value string) string {
...
+				fmt.Fprintf(cmd.OutOrStdout(), "Name for this credential [%s]: ", defaultName)
+				input, err := reader.ReadString('\n')
+				if err != nil {
+					return fmt.Errorf("reading name: %w", err)
+				}
+				name = strings.TrimSpace(input)
+				if name == "" {
+					name = defaultName
+				}
+			}
```
dbmikus, line 492: make a `parseClaudeName` function for this

internal/auth/discovery.go
```diff
@@ -567,6 +567,88 @@ func parseEpochMillis(raw any) (int64, error) {
...
+	Type string
+	// Source describes where the credential was found (e.g. file path or "macOS Keychain").
+	Source string
+	// Value is the raw credential data to upload: the API key string or the full OAuth JSON.
+	Value string
+}
+
+// DiscoverClaudeCredentials scans local credential sources and returns all
+// Claude-specific credentials found, preserving their type and source.
+func DiscoverClaudeCredentials(homeDir string) ([]ClaudeCredential, error) {
```
dbmikus, line 583: State the comment to say if the user specifies an empty home directory, we get it from their operating system or environment variables.

cmd/amika/secrets.go
```diff
@@ -398,12 +403,340 @@ func maskValue(value string) string {
...
+			fmt.Fprintf(cmd.OutOrStdout(), "Deleted credential %s\n", args[0])
+			return nil
+		},
+	}
+}
+
+// autoResolveClaudeCredential resolves a credential value automatically based
+// on the requested type, without interactive prompts.
+//   - "api_key": reads the ANTHROPIC_API_KEY environment variable.
+//   - "oauth": on macOS reads from the keychain, otherwise from credential files.
```
dbmikus, line 667: Does our command help text describe how we auto resolve these credential types? I would want to know from what environment variables or files or keychains we're pulling this data from.

cmd/amika/secrets.go
```diff
@@ -398,12 +403,340 @@ func maskValue(value string) string {
...
+	if credType == "api_key" {
+		key := os.Getenv("ANTHROPIC_API_KEY")
+		if key == "" {
+			return "", fmt.Errorf("ANTHROPIC_API_KEY environment variable is not set")
+		}
+		return key, nil
+	}
+
+	// OAuth: try keychain on macOS, then credential files.
+	if runtime.GOOS == "darwin" {
```
dbmikus, line 678: Can we make this a function that is just called `isMacOS`, and then we can use that function in other places?

cmd/amika/secrets.go
```diff
@@ -398,12 +403,340 @@ func maskValue(value string) string {
...
+
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("determining home directory: %w", err)
+	}
+
+	oauthPaths := []string{
+		filepath.Join(homeDir, ".claude", ".credentials.json"),
+		filepath.Join(homeDir, ".claude-oauth-credentials.json"),
+	}
```
dbmikus, line 693: Do we have similar code anywhere in our `amika secret extract` flows? At the least, we should have a consistent way of getting these OAuth file paths.

cmd/amika/secrets_test.go
dbmikus, line 1: For all the tests that use our API client, what API URL are they hitting? I don't want tests to actually call a production API.

internal/auth/discovery.go
```diff
@@ -567,6 +567,88 @@ func parseEpochMillis(raw any) (int64, error) {
...
+		}
+	}
+
+	var creds []ClaudeCredential
+
+	// Check for API keys in Claude config files.
+	apiKeyPaths := []string{
+		filepath.Join(homeDir, ".claude.json.api"),
+		filepath.Join(homeDir, ".claude.json"),
+	}
```
dbmikus, line 598: This looks like another place that we can consolidate these Claude file paths.

internal/auth/discovery.go
```diff
@@ -567,6 +567,88 @@ func parseEpochMillis(raw any) (int64, error) {
...
+				break // one API key per file is enough
+			}
+		}
+	}
+
+	// Check for OAuth credentials in files.
+	oauthPaths := []string{
+		filepath.Join(homeDir, ".claude", ".credentials.json"),
+		filepath.Join(homeDir, ".claude-oauth-credentials.json"),
+	}
```
dbmikus, line 626: Consolidate the Claude file paths. Maybe just use constants or a function or something.

dbmikus (commented):
Can we rename this to `amika secret claude push` for consistency with `amika secret push`? It's doing the same thing with custom Claude handling.

Also, this overlaps with `amika secret extract --push`. We probably want to consolidate them, eventually. Separate command is fine for now. I'd rather consolidate things when we have time think harder about the interface.

PR #120, opened 2026-03-28
Address PR #119 code review feedback

dbmikus:
## Summary

- Rename `secret claude upload` to `secret claude push` for consistency with `amika secret push`
- Fix misplaced doc comments (`resolveGitURL` comment was above `autoSelectClaudeCredential`)
- Expand push command help text to document credential auto-resolution sources
- Extract `parseClaudeCreds` and `parseClaudeName` functions with explicit mutual exclusivity check for `--value`/`--from-file`
- Consolidate hardcoded Claude credential file paths into shared `ClaudeAPIKeyPaths()` and `ClaudeOAuthPaths()` helpers in `internal/auth`
- Add `isMacOS()` helper to replace inline `runtime.GOOS == "darwin"` checks
- Replace tests that relied on auth failures (hitting production) with `httptest` mock server

### DIFF
--- a/cmd/amika/secrets.go
+++ b/cmd/amika/secrets.go
@@ -619,7 +619,7 @@ func discoverAllClaudeCredentials() ([]auth.ClaudeCredential, error) {
 	}
 
 	// On macOS, also try the keychain.
-	if runtime.GOOS == "darwin" {
+	if isMacOS() {
 		keychainValue, err := readClaudeCredentialFromKeychain()
 		if err == nil && keychainValue != "" && json.Valid([]byte(keychainValue)) {
 			creds = append(creds, auth.ClaudeCredential{
@@ -713,7 +713,7 @@ func autoResolveClaudeCredential(credType string) (string, error) {
 	}
 
 	// OAuth: try keychain on macOS, then credential files.
-	if runtime.GOOS == "darwin" {
+	if isMacOS() {
 		value, err := readClaudeCredentialFromKeychain()
 		if err == nil && value != "" {
 			return value, nil
@@ -741,6 +741,11 @@ func autoResolveClaudeCredential(credType string) (string, error) {
 	}
 
 	return "", fmt.Errorf("no OAuth credentials found; on macOS check keychain, or provide --value or --from-file")
+}
+
+// isMacOS reports whether the current operating system is macOS.
+func isMacOS() bool {
+	return runtime.GOOS == "darwin"
 }
 
 // claudeCredentialTypeToAPI maps the discovery type label to the API type field.
--- a/cmd/amika/secrets.go
+++ b/cmd/amika/secrets.go
@@ -725,9 +725,9 @@ func autoResolveClaudeCredential(credType string) (string, error) {
 		return "", fmt.Errorf("determining home directory: %w", err)
 	}
 
-	oauthPaths := []string{
-		filepath.Join(homeDir, ".claude", ".credentials.json"),
-		filepath.Join(homeDir, ".claude-oauth-credentials.json"),
+	oauthPaths := make([]string, len(auth.ClaudeOAuthPaths()))
+	for i, p := range auth.ClaudeOAuthPaths() {
+		oauthPaths[i] = filepath.Join(homeDir, p)
 	}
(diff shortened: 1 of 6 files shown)


cmd/amika/secrets.go
```diff
@@ -530,6 +478,84 @@ Examples:
 	return cmd
 }
 
+// parseClaudeCreds resolves the credential value and type from command flags.
+// It returns an error if mutually exclusive flags --value and --from-file are
+// both provided.
```
dbmikus: Your comment should describe the two strings that are returned, because otherwise I don't know what either of these return strings is. I should just be able to read the comment and understand: what do I pass into this function? What does it do? What does it return?

cmd/amika/secrets.go
```diff
@@ -743,6 +743,11 @@ func autoResolveClaudeCredential(credType string) (string, error) {
 	return "", fmt.Errorf("no OAuth credentials found; on macOS check keychain, or provide --value or --from-file")
 }
 
+// isMacOS reports whether the current operating system is macOS.
+func isMacOS() bool {
+	return runtime.GOOS == "darwin"
+}
+
```
dbmikus: Put this this inside an `internal/arch` package.

cmd/amika/secrets_test.go
```diff
@@ -400,20 +403,20 @@ func TestClaudeCredentialTypeToAPI(t *testing.T) {
...
 	cmd := exec.Command(bin, "secret", "claude", "push",
 		"--value", `{"claudeAiOauth":{"accessToken":"test"}}`,
 		"--name", "Test OAuth")
+	cmd.Env = withEnv(os.Environ(), mockAPIEnv(srv.URL)...)
 	out, err := cmd.CombinedOutput()
-	// Expect failure due to no auth, but NOT a validation error.
-	if err == nil {
-		t.Fatalf("expected error (no auth), got success\noutput:\n%s", out)
+	if err != nil {
+		t.Fatalf("unexpected error:\n%s", out)
```
dbmikus: Why are you redoing the error `t.Fatalf` error messages in here?

cmd/amika/secrets_test.go
```diff
@@ -400,20 +403,20 @@ func TestClaudeCredentialTypeToAPI(t *testing.T) {
 
 func TestSecretClaudePush_ValueFlag(t *testing.T) {
 	bin := buildAmika(t)
+	srv := newMockClaudeAPI(t)
+	defer srv.Close()
```
dbmikus: Since we pass in `t` to `newMockClaudeAPI`, Is there a way to make it do `srv.close` automatically when the test is done so that we don't need to remember to call the `defer srv.close`?

cmd/amika/secrets_test.go
```diff
@@ -400,20 +403,20 @@ func TestClaudeCredentialTypeToAPI(t *testing.T) {
...
 
-	// With --value and --name, the command will try to hit the API and fail,
-	// but we can verify it gets past validation.
 	cmd := exec.Command(bin, "secret", "claude", "push",
 		"--value", `{"claudeAiOauth":{"accessToken":"test"}}`,
 		"--name", "Test OAuth")
+	cmd.Env = withEnv(os.Environ(), mockAPIEnv(srv.URL)...)
```
dbmikus: I don't want the user to have to remember to always extend the environment to include SRV_URL. Can we have some function that first calls `newMockClaudeAPI` and then it modifies the `os.Environ` for the duration of the test? Or would it be better to have it modify the `cmd.Env`?

cmd/amika/secrets_test.go
```diff
@@ -594,6 +603,38 @@ func writeJSONFixture(t *testing.T, path string, content string) {
 	}
 }
 
+// newMockClaudeAPI starts an httptest server that handles Claude secret API
+// endpoints and returns it. The caller should defer server.Close().
+func newMockClaudeAPI(t *testing.T) *httptest.Server {
```
dbmikus: Can you add a TODO comment to extract this functionality out to somewhere we can use it across all tests?

jdc123 (approved):
lgtm

PR #121, opened 2026-03-28
feat: auto-select Claude credential with OAuth preference

dbmikus:
Prefer the first OAuth credential when multiple exist on the remote. Fall back to the first API key if no OAuth credentials are found. Print the selected credential name and type to stderr.


jdc123 (approved):
lgtm

PR #122, opened 2026-03-28
fix: remove duplicate function declarations breaking build

dbmikus:
Remove duplicate declarations of StartDockerSandbox, Client.StartSandbox, and sandboxStartCmd that were introduced by a merge artifact.


dbmikus:
I had two LLMs double-check line-by-line that there were no differences in the duplicate functions we removed.

PR #124, opened 2026-03-30
feat: add `sandbox agent-send` command for sending messages to sandbox agents

jdc123:
Add a new `amika sandbox agent-send <name> [message]` CLI command that sends prompts to AI agent CLIs (Claude, Codex, etc.) running inside sandbox containers.

Two execution modes:
- **Wait (default):** Streams agent stdout/stderr in real-time via `docker exec` (local) or SSH (remote), blocking until the agent finishes.
- **No-wait (`--no-wait`):** Launches the agent in a detached tmux session so the process survives SSH/container disconnect, then returns immediately.

The command resolves messages from positional args or piped stdin, tries the local sandbox store first, and falls back to remote SSH — matching the existing `sandbox connect` pattern.

Agent CLI differences are handled via a config map (`knownAgents`) that maps agent names to their binary and non-interactive print flag (e.g. claude → `-p`, codex → `--prompt`). Unknown agents default to the name as-is with `-p`.

Flags: --no-wait, --agent (default "claude"), --workdir (default /home/amika)

### DIFF
--- a/cmd/amika/sandbox.go
+++ b/cmd/amika/sandbox.go
@@ -2009,15 +2009,19 @@ var knownAgents = map[string]agentConfig{
 }
 
 // resolveAgentConfig returns the agent configuration for the given name.
-// Unknown agents use the name as the binary with "-p" as the default flag.
-func resolveAgentConfig(name string) agentConfig {
+// Returns an error if the agent is not in the knownAgents map.
+func resolveAgentConfig(name string) (agentConfig, error) {
 	if cfg, ok := knownAgents[name]; ok {
-		return cfg
+		return cfg, nil
 	}
-	return agentConfig{Binary: name, PrintArg: "-p"}
+	known := make([]string, 0, len(knownAgents))
+	for k := range knownAgents {
+		known = append(known, fmt.Sprintf("%q", k))
+	}
+	return agentConfig{}, fmt.Errorf("unknown agent %q; supported agents: %s", name, strings.Join(known, ", "))
 }
 
-var runSandboxAgentSend = func(name, message string, noWait bool, workdir string, agent agentConfig, stdout, stderr io.Writer) error {
+func runDockerSandboxAgentSend(name, message string, noWait bool, workdir string, agent agentConfig, stdout, stderr io.Writer) error {
 	dockerArgs := buildSandboxAgentSendArgs(name, message, noWait, workdir, agent)
 	dockerCmd := exec.Command("docker", dockerArgs...)
 	if !noWait {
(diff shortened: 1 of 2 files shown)


cmd/amika/sandbox.go
```diff
@@ -1996,6 +1996,150 @@ Examples:
...
+	"claude": {Binary: "claude", PrintArg: "-p", ExtraArgs: []string{"--dangerously-skip-permissions"}},
+}
+
+// resolveAgentConfig returns the agent configuration for the given name.
+// Unknown agents use the name as the binary with "-p" as the default flag.
+func resolveAgentConfig(name string) agentConfig {
+	if cfg, ok := knownAgents[name]; ok {
+		return cfg
+	}
+	return agentConfig{Binary: name, PrintArg: "-p"}
```
dbmikus: Should we instead return an error here? I feel like that's more debuggable than defaulting to the agent name and `-p`, and it also accounts for cases when `-p` might be an agent argument, but with the wrong behavior.

cmd/amika/sandbox.go
```diff
@@ -1996,6 +1996,150 @@ Examples:
...
+}
+
+var runSandboxAgentSend = func(name, message string, noWait bool, workdir string, agent agentConfig, stdout, stderr io.Writer) error {
```
dbmikus: nit: why do we do `var runSandboxAgentSend = func (...)` instead of `func sandboxAgentSend`?

And not a nit: if this is for Docker, can we call it `runDockerSandboxAgentSend`?

cmd/amika/sandbox.go
```diff
@@ -1996,6 +1996,150 @@ Examples:
...
+		}
+
+		agentStr := strings.Join(agentCmdParts(agent, fmt.Sprintf("%q", message)), " ")
+
+		if noWait {
+			// Run in a detached tmux session so the agent survives SSH disconnect.
+			// Pass as a single string so the remote shell keeps the tmux command intact.
+			sessionName := fmt.Sprintf("amika-agent-send-%d", time.Now().UnixNano())
+			shellCmd := fmt.Sprintf("tmux new-session -d -s '%s' 'cd %s && %s'",
+				sessionName, workdir, agentStr)
```
dbmikus: Why do we have this tmux session creation stuff here, and also above at line 2042? I think we should use `buildSandboxAgentSendArgs` here.

dbmikus (commented):
You want the default workdir to be `AMIKA_AGENT_CWD` environment variable, which makes sure this runs from within the repo instead of the home directory.

cmd/amika/sandbox.go
```diff
@@ -2280,6 +2277,6 @@ func init() {
 	sandboxSSHCmd.Flags().Bool("revoke", false, "Revoke SSH access for the sandbox")
 	sandboxCodeCmd.Flags().String("editor", "cursor", "Editor to open (currently only \"cursor\" is supported)")
 	sandboxAgentSendCmd.Flags().Bool("no-wait", false, "Send the instruction and return immediately without waiting for a response")
-	sandboxAgentSendCmd.Flags().String("workdir", sandboxConnectWorkdir, "Working directory inside the container")
+	sandboxAgentSendCmd.Flags().String("workdir", "$AMIKA_AGENT_CWD", "Working directory inside the container (default: $AMIKA_AGENT_CWD)")
```
dbmikus, line 2280: Does this evaluate `$AMIKA_AGENT_CWD` locally or on the sandbox?

dbmikus (approved):

PR #126, opened 2026-03-30
fix: set CWD to agent working directory before running setup.sh

dbmikus:
## Summary

- `run-hook.sh` now reads `/var/lib/amikad/agent-cwd` (written by `pre-setup.sh`) and `cd`s into it before executing `setup.sh`, so the user-provided setup hook runs from the project directory instead of `/home/amika`.
- Only `setup.sh` is affected — `pre-setup.sh` and `post-setup.sh` keep their existing CWD behavior.
- Extracts `RequireExpensiveDockerTests` into `test/testutil` to deduplicate the skip check across test packages.

### DIFF
--- a/cmd/amika/materialize_test.go
+++ b/cmd/amika/materialize_test.go
@@ -8,10 +8,9 @@ import (
 	"strings"
 	"testing"
 	"time"
-)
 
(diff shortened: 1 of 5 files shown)


jdc123 (approved):
lgtm

PR #127, opened 2026-03-30
feat: add `--branch` flag to sandbox create for git branch support

jdc123:
Pass a branch argument through the CLI, public service API, HTTP API, and remote API client so sandboxes can clone a specific git branch instead of the repo default. The branch is persisted in local sandbox state and displayed in `sandbox list`.

### DIFF
--- a/cmd/amika/sandbox.go
+++ b/cmd/amika/sandbox.go
@@ -163,9 +163,11 @@ var sandboxCreateCmd = &cobra.Command{
 		}
 		image = resolvedImage.Image
 
+		branchFlag, _ := cmd.Flags().GetString("branch")
 		collected, err := collectMounts(mountStrs, volumeStrs, portStrs, portHostIP,
 			gitPath, gitFlagChanged, noClean,
-			setupScript, cmd.Flags().Changed("setup-script"))
+			setupScript, cmd.Flags().Changed("setup-script"),
+			branchFlag)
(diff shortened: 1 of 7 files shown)


cmd/amika/sandbox.go
```diff
@@ -1109,8 +1114,13 @@ func resolveGitRoot(startPath string) (string, error) {
 	return "", fmt.Errorf("no git repository root found from %q", absPath)
 }
 
-func cloneGitRepo(src, dst string) error {
-	cmd := exec.Command("git", "clone", "--local", "--no-hardlinks", src, dst)
+func cloneGitRepo(src, dst, branch string) error {
+	args := []string{"clone", "--local", "--no-hardlinks"}
```
dbmikus, line 1118: Can you add this comment here? I wasn't sure what these extra args were doing (even though I added them originally) and I was concerned that the `--local` flag would cause errors if we were cloning from github.

```
// Clone with --local --no-hardlinks: when the source is a local path,
// --local skips the git transport protocol for speed, and --no-hardlinks
// forces a full copy of objects instead of hardlinking back to the source
// repo, so the sandbox can't corrupt the user's real repo.
// Both flags are silently ignored for remote URLs.
```

dbmikus (approved):
Looks good to me. Had one small request to add a comment.

PR #128, opened 2026-03-30
fix: resolve two test-expensive failures

dbmikus:
Fix pre-setup.sh pnpm directory creation failing when Docker bind-mounts (e.g. agent credential files under ~/.local/) leave intermediate directories root-owned. Create as root then chown instead of running mkdir as the amika user.

Fix sandbox lifecycle integration test failing because sandbox delete prompts for confirmation and receives EOF. Pass --force to skip the prompt.


jdc123 (approved):
LGTM

PR #134, opened 2026-04-01
feat: add session management to `sandbox agent-send` for remote sandboxes

jdc123:
By default agent-send now resumes the latest active session so multi-turn conversations persist without manual session tracking. Use --new-session to start fresh, --session-id to resume a specific session, or --resume to continue the most recent one.

Key changes:
- Add session CRUD methods to apiclient (create, list, get, update)
- Replace syscall.Exec with exec.Command on the remote path so stdout can be captured and parsed for the Claude session ID
- Use --output-format json in wait mode; extract session_id from the response and store it via POST /api/sandboxes/:id/sessions
- Map amika flags to Claude CLI flags correctly (SessionID → --resume, Resume → --continue, NewSession → no flag)
- Ignore SSH exit 255 when valid JSON output was already received
- Silence duplicate Cobra error output on agent-send

### DIFF
--- a/cmd/amika/sandbox.go
+++ b/cmd/amika/sandbox.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"bufio"
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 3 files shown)


cmd/amika/sandbox.go
```diff
@@ -2124,6 +2100,116 @@ func buildAgentShellCmd(message string, noWait bool, workdir string, agent agent
...
+// via the API, and prints the result text to stdout.
+func runRemoteAgentSend(client *apiclient.Client, name, message string, noWait bool, workdir string, agent agentConfig, opts agentRunOpts, stdout, stderr io.Writer) error {
+	// Resolve session for default behavior (no explicit session flags).
+	var existingSession *apiclient.Session
+	if !opts.NewSession && opts.SessionID == "" && !opts.Resume {
+		session, err := client.GetLatestSession(name)
+		if err != nil {
+			fmt.Fprintf(stderr, "Warning: could not look up latest session: %v\n", err)
+		} else if session != nil && session.Status == "active" {
+			if claudeID, ok := session.Metadata["claude_session_id"].(string); ok && claudeID != "" {
```
dbmikus, line 2134: Can we make a follow up ticket in Linear, to segment `session.Metadata` and `session.SystemMetadata`? I remember shoving system-internal hacks into the user-facing metadata and it ended up leading to headaches later. Our metadata and metadata that is settable by clients should be in different namespaces.

Then can we add a TODO comment here to fix this?

cmd/amika/sandbox.go
```diff
@@ -2124,6 +2100,116 @@ func buildAgentShellCmd(message string, noWait bool, workdir string, agent agent
...
+		}
+
+		// If we got valid, non-error JSON output, ignore SSH-level exit
+		// codes (e.g. 255 from connection teardown).
+		if !parsed.IsError {
+			return nil
+		}
+	} else if len(output) > 0 {
+		// Could not parse JSON — print raw output.
+		stdout.Write(output) //nolint:errcheck
```
dbmikus: Don't we also want to return an error here? Like, being unable to parse the JSON is a CLI error right?

cmd/amika/sandbox.go
```diff
@@ -2124,6 +2100,116 @@ func buildAgentShellCmd(message string, noWait bool, workdir string, agent agent
...
+		fmt.Fprint(stdout, parsed.Result)
+		if parsed.Result != "" && !strings.HasSuffix(parsed.Result, "\n") {
+			fmt.Fprintln(stdout)
+		}
+
+		// Store session if this was a new session (no existing session being resumed).
+		if parsed.SessionID != "" && existingSession == nil {
+			if _, createErr := client.CreateSession(name, apiclient.CreateSessionRequest{
+				AgentName: agent.Binary,
+				Metadata: map[string]interface{}{
```
dbmikus, line 2184: This is the other place we should swap to `SystemMetadata`

dbmikus (approved):
LGTM

PR #135, opened 2026-04-06
Default sandbox commands to remote mode

dbmikus:
## Summary
- **Default mode is now `remote`** for all sandbox commands, regardless of login status. Previously, logged-out users silently fell back to local-only mode.
- **Explicit `--local` flag required** to operate on local sandboxes. The `"both"` (local+remote) mode is eliminated — each command targets exactly one provider.
- **Clear auth errors** when remote mode is used without authentication: `not logged in; run "amika auth login" or use --local"`. The `AMIKA_API_KEY` env var continues to bypass session auth.
- **New `internal/runmode` package** extracts mode resolution and auth guard logic out of the large `sandbox.go`, with full unit test coverage.
- **connect and agent-send** now respect `--local`/`--remote` flags (previously they ignored them and always tried local first).
- **ssh and code** commands now use the mode system instead of manually checking the local store.

## Test plan
- [x] `go test ./internal/runmode/...` — new unit tests for `Resolve` and `RequireAuth`
- [x] `go test ./cmd/amika/...` — existing tests pass (updated `TestSandboxListCommand_PrintsRows` to use `--local`)
- [x] `make build` — compiles cleanly
- [x] Manual: `amika sandbox list` while logged out → auth error
- [x] Manual: `amika sandbox list --local` while logged out → shows local sandboxes
- [x] Manual: `amika sandbox list` while logged in → shows remote sandboxes

### DIFF
--- a/cmd/amika/sandbox.go
+++ b/cmd/amika/sandbox.go
@@ -21,6 +21,7 @@ import (
 	"github.com/gofixpoint/amika/internal/auth"
 	"github.com/gofixpoint/amika/internal/config"
 	"github.com/gofixpoint/amika/internal/constants"
+	"github.com/gofixpoint/amika/internal/runmode"
 	"github.com/gofixpoint/amika/internal/sandbox"
 	"github.com/gofixpoint/amika/internal/txn"
 	"github.com/gofixpoint/amika/pkg/amika"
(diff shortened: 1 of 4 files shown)


jdc123 (approved):
LGTM

PR #136, opened 2026-04-06
Add .bashrc and set TERM=xterm in sandbox shell configs

dbmikus:
## Summary
- Add a `.bashrc` to sandbox containers that mirrors the existing `.zshrc` (prompt, history, completion, color, pnpm PATH)
- Set `export TERM=xterm` in both `.zshrc` and `.bashrc` so child processes (less, git, editors) render correctly
- Add build context extraction test for `.bashrc`

## Test plan
- [x] `TestWritePresetBuildContext_ExtractsBashrc` passes
- [x] `TestWritePresetBuildContext_ExtractsZshrc` still passes
- [x] `make fmt`, `make vet`, `make lint` all pass
- [x] Rebuild a preset image and verify bash/zsh sessions have the prompt, history, and `TERM=xterm`

### DIFF
--- a/internal/sandbox/presets/.bashrc
+++ b/internal/sandbox/presets/.bashrc
@@ -1,3 +1,6 @@
+# Set terminal type for proper rendering in tools like less, git, etc.
+export TERM=xterm
+
 # Set up the prompt (multi-line, similar to zsh adam1 theme)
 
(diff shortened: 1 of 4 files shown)


jdc123 (approved):
lgtm

PR #137, opened 2026-04-07
feat: poll for status after remote sandbox start/stop

jdc123:
The API now returns 202 for start/stop and transitions through intermediate states. Add WaitForSandboxStart/WaitForSandboxStop polling methods and wire them into the CLI commands to match the create-command pattern. Also remove duplicate sandboxStartCmd registration.

### DIFF
--- a/internal/apiclient/client.go
+++ b/internal/apiclient/client.go
@@ -88,22 +88,30 @@ func (c *Client) GetSandbox(name string) (*RemoteSandbox, error) {
 	return &result, nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/apiclient/client.go
dbmikus, line 1: nit: can you dedupe the logic in:

- `WaitForSandbox`
- `WaitForSandboxStart`
- `WaitForSandboxStop`

AFAIK they all do the same thing, but just poll for different statuses. It's useful to have one implementation so if we change how we poll for sandbox status we only need to make one update.

PR #138, opened 2026-04-08
Add "ls" and "rm" aliases to list and delete subcommands

dbmikus:
## Summary
- Add `ls` alias to all `list` subcommands (sandbox, volume, service, secret claude)
- Add `rm` alias to `secret claude delete` (sandbox delete and volume delete already had it)

## Test plan
- [x] `amika sandbox ls` works like `amika sandbox list`
- [x] `amika volume ls` works like `amika volume list`
- [x] `amika service ls` works like `amika service list`
- [x] `amika secret claude ls` works like `amika secret claude list`
- [x] `amika secret claude rm` works like `amika secret claude delete`


jdc123 (approved):
lgtm

PR #139, opened 2026-04-08
URL-encode sandbox names in API client

dbmikus:
## Summary
- Add `url.PathEscape()` to all sandbox name and session ID interpolations in `internal/apiclient/client.go`
- Add unit tests verifying that names with slashes (e.g. `org/proj`) produce properly encoded paths (`org%2Fproj`)

Sandbox names containing `/` were being split into multiple URL path segments, causing 404 errors on get, delete, start, stop, SSH, and session operations.

## Test plan
- [x] `go test ./internal/apiclient/...` passes
- [x] Existing tests still pass
- [x] Manual: create sandbox with `/` in name, verify get/delete/start/stop work from CLI

### DIFF
--- a/internal/apiclient/client.go
+++ b/internal/apiclient/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


jdc123 (approved):
lgtm

PR #147, opened 2026-04-10
fix: improve agent-send error message for authentication failures

jdc123:
## Summary
- When the sandbox agent fails to authenticate with its AI provider (e.g. Anthropic 401), the error was an opaque HTTP 500 with deeply nested JSON that was hard to read
- Now the client detects authentication failures in the agent result and returns a clear message: explains the agent failed to authenticate with its AI provider and that credentials may have expired or been revoked
- Introduces an `HTTPError` type in `apiclient` so callers can inspect HTTP status codes and response bodies from `doJSON` errors

**Before:**
```
agent-send failed for sandbox "amber-heron": remote agent-send: HTTP 500: {"error":"Agent command failed","details":"{\"type\":\"result\",\"subtype\":\"success\",\"is_error\":true,...}"}
```

**After:**
```
agent-send failed for sandbox "amber-heron": remote agent-send: agent failed to authenticate with its AI provider: Failed to authenticate. API Error: 401 {"type":"error","error":{"type":"authentication_error","message":"Invalid authentication credentials"}}

the sandbox agent's API credentials may have expired or been revoked; recreate the sandbox or update its API keys to restore access
```

## Test plan
- [x] Added unit tests for `extractAgentAuthError` covering: non-HTTP errors, non-JSON bodies, auth failures, non-auth agent errors, and `is_error=false`
- [x] Added integration-style test for `AgentSend` with a mock server returning a 500 auth failure
- [x] All existing tests pass
- [x] Lint, vet, fmt all pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/apiclient/client.go
+++ b/internal/apiclient/client.go
@@ -4,6 +4,7 @@ package apiclient
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -11,6 +12,18 @@ import (
 	"strings"
 	"time"
 )
+
+// HTTPError is returned by doJSON when the server responds with a non-2xx
+// status code. It carries the raw status and body so callers can inspect or
+// parse the response for structured error information.
+type HTTPError struct {
+	StatusCode int
+	Body       string
+}
+
+func (e *HTTPError) Error() string {
+	return fmt.Sprintf("HTTP %d: %s", e.StatusCode, e.Body)
+}
 
 // Client calls the remote Amika API with a bearer token.
 type Client struct {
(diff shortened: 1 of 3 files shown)


internal/apiclient/client.go
dbmikus, line 1: code organization-wise: can we put the error stuff inside `internal/apiclient/errors.go`? It will be part of the same `apiclient` package, but then we can separate and streamline all our error functions.

dbmikus (approved):
LGTM, but would you mind isolating the error code inside a new file in the `apiclient` package?

jdc123:
@codex review

PR #149, opened 2026-04-14
feat: add codex support to agent-send command

jdc123:
Extend agentConfig to support agents with subcommand-based CLIs (like Codex's `exec` / `exec resume`) alongside flag-based CLIs (like Claude). Add "codex" to knownAgents so `--agent codex` works for agent-send.


dbmikus (approved):
FYI, I think we should start incrementally factoring things out of this giant sandbox.go file instead of just layering more stuff on it.

But for this PR, no need. Go and merge. LGTM

PR #153, opened 2026-04-16
Fix sandbox terminal rendering for Claude CLI

dbmikus:
## Summary

- **Set `LANG=C.UTF-8` in base Dockerfile** — ensures all processes inherit a UTF-8 locale instead of falling back to POSIX, which was causing garbled Unicode characters in TUI apps like `claude` inside tmux.
- **Vendor and install Ghostty terminfo** — self-contained terminfo source from `ghostty-org/ghostty` compiled into the system terminfo database so `TERM=xterm-ghostty` works over SSH from Ghostty without fallback.
- **Add `.tmux.conf`** — sensible defaults: true-color passthrough for Ghostty/xterm-256color/alacritty, low escape-time for vim, mouse support, focus events, 10k scrollback.
- **Refine TERM guard in `.zshrc`/`.bashrc`** — only upgrade truly basic terminals (`linux`, `dumb`); capable terminals like Ghostty keep their native TERM, and tmux keeps `tmux-256color`.

## Context

When running `claude` CLI inside a sandbox tmux session (especially over SSH from Ghostty), box-drawing and other Unicode characters render incorrectly. Root causes:
1. Locale defaults to POSIX because nothing in the image sets `LANG`.
2. `.zshrc`/`.bashrc` unconditionally override `TERM=xterm`, clobbering tmux's correct `TERM=tmux-256color`.
3. No Ghostty terminfo on the remote, so SSH from Ghostty falls back to generic xterm.
4. No tmux config, so no true-color passthrough or other modern terminal features.

## Test plan

- [x] Build the base Docker image and verify `locale` shows `C.UTF-8`
- [x] Verify `infocmp xterm-ghostty` works inside the container
- [x] Open a tmux session and verify `echo $TERM` shows `tmux-256color`
- [x] SSH from Ghostty without tmux and verify `echo $TERM` shows `xterm-ghostty`
- [x] Run `claude` CLI inside tmux and confirm characters render correctly


jdc123 (approved):
lgtm

PR #169, opened 2026-04-22
auth: support WorkOS organization API key as a login credential

jdc123:
Add `amika auth login --api-key-file <path>` (and `-` for stdin) so CI and headless environments can authenticate without the interactive WorkOS device flow. The key is persisted at `~/.local/state/amika/api-key.json` (0600) alongside the existing session file.

Credential precedence is now resolved in one place via `apiclient.NewResolvedTokenSource`, re-evaluated on every request:

  1. AMIKA_API_KEY environment variable
  2. Stored API key file
  3. WorkOS session (with auto-refresh)

`amika auth login` refuses to stack credentials — callers must `amika auth logout` first. `amika auth logout` now clears both the stored API key and the WorkOS session. `amika auth status` reports the winning source and explicitly calls out shadowed sources.

### DIFF
--- a/cmd/amika/auth.go
+++ b/cmd/amika/auth.go
@@ -1,8 +1,11 @@
 package main
 
 import (
+	"errors"
 	"fmt"
+	"io"
 	"os"
+	"time"
 
 	"github.com/gofixpoint/amika/internal/auth"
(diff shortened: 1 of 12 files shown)


dbmikus (approved):
LGTM

Issue #157, opened 2026-04-17
Allow modifying sandbox metadata after creation


dbmikus:
## Context

Once a sandbox is created, users cannot update its properties via the CLI. This forces users to destroy and recreate sandboxes just to change a name or adjust timeouts.

## Proposed solution

Support post-creation modification of sandbox metadata via CLI:
- Rename sandboxes
- Modify TTL
- Set inactivity timeout
- Set auto-delete timeout for suspended sandboxes

PR #178, opened 2026-04-25
feat: allow modifying sandbox metadata after creation

Abhijitam01:
Add CLI sandbox update command and HTTP PATCH endpoint to modify sandbox metadata including rename, TTL, and timeout settings. Extends Service interface with UpdateSandbox method using immutable update pattern with pointer-based optional fields for selective updates. Implements Docker container rename integration and duration validation. Updates SandboxRecord with TTL, InactivityTimeout, and AutoDeleteTimeout fields persisted as JSONL with omitempty tags for backward compatibility.

closses #157


dbmikus:
Hey, just wanted to say thanks for your contributions. We'll get to reviewing them! Underwater with a few other things right now.

PR #197, opened 2026-05-15
Mirror start.sh log alongside setup.sh in run-hook

jdc123:
## Summary
- Extend `run-hook.sh` to treat `start.sh` like `setup.sh`: log to `/var/log/amika/start.log` (writable by the `amika` user) and mirror to `/var/log/amikad/start.log` when the hook finishes.
- Parameterize the log path off `${script_name%.sh}` so the user-facing branch handles both hooks without duplication.
- Update `TestPresetRunHook_*` to assert the broadened condition and parameterized path; rename to `…ForUserHooksAndMirrorsThemToAmikad`.

## Why
The control plane reads `/var/log/amikad/setup.log` to surface setup-script failures back to the user. A `start.sh` hook is materialized at `/usr/local/etc/amikad/setup/start.sh` (see `SANDBOX_START_SCRIPT` in `amika-mono`), and it needs the same log treatment so failures during the start phase can be surfaced the same way.

## Test plan
- [x] `make fmt`, `make vet`, `make test-unit` clean.
- [x] `go test ./internal/sandbox -run TestPresetRunHook -v` passes.
- [ ] Manual: invoke `run-hook.sh /usr/local/etc/amikad/setup/start.sh` inside a built image; confirm `/var/log/amika/start.log` is created and mirrored to `/var/log/amikad/start.log`.

### DIFF
--- a/internal/sandbox/pre_setup_test.go
+++ b/internal/sandbox/pre_setup_test.go
@@ -98,7 +98,7 @@ func TestPresetPreSetup_OpenCodeGatingContract(t *testing.T) {
 	}
 }
 
-func TestPresetRunHook_UsesAmikaOnlyForSetupAndMirrorsItToAmikad(t *testing.T) {
+func TestPresetRunHook_UsesAmikaForUserHooksAndMirrorsThemToAmikad(t *testing.T) {
 	data, err := presetFS.ReadFile("presets/run-hook.sh")
 	if err != nil {
 		t.Fatalf("ReadFile failed: %v", err)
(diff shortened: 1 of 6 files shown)


bin/build-docker-images
```diff
@@ -3,7 +3,7 @@ set -euo pipefail
 
 #/ Usage: bin/build-docker-images <image>... [--platform <platform>] [--ignore-unstaged] [--push-daytona] [--ci]
 #/
-#/ Build preset Docker images and tag them with the current git commit.
+#/ Build preset Docker pimages and tag them with the current git commit.
```
dbmikus: weird little comment bug?

dbmikus (approved):
LGTM

PR #224, opened 2026-06-02
Add `amika sessions capture-init` to mirror agent transcripts

jdc123:
## Summary

- Add `amika sessions capture-init`, which installs hooks into Claude Code and Codex so each agent mirrors its session transcripts into the amika state directory on every turn — no daemon.
- Add the hook-invoked `amika sessions capture --source {claude|codex}` that does the actual copy into `<state>/sessions/{claude,codex}/<YYYY-MM-DD>/`, grouping a day's sessions together.
- Idempotent: existing Claude hook events and Codex TOML keys / comments are preserved; an existing non-amika `notify` is left alone with a warning.

## How it works

`amika sessions capture-init`:
- Appends a `Stop` hook to `~/.claude/settings.json` running `amika sessions capture --source claude`. Other keys (theme, other hook events) round-trip unchanged.
- Sets a top-level `notify = [...]` array in `~/.codex/config.toml` running `amika sessions capture --source codex`. The TOML is edited line-wise so existing formatting / comments survive, and the assignment is inserted before the first `[section]` header.

`amika sessions capture --source ...` (invoked by the agent hooks):
- Claude: reads the `transcript_path` from the JSON payload Claude pipes on stdin and atomically copies it to `<state>/sessions/claude/<YYYY-MM-DD>/<session_id>.jsonl`, where the day stamp comes from the transcript's mtime.
- Codex: Codex's notify payload doesn't include a session path, so we walk `<codex-home>/sessions/**` and mirror every `.jsonl` whose source mtime is newer than its mirror (or that has no mirror yet) into `<state>/sessions/codex/<YYYY-MM-DD>/`. The day stamp is derived from Codex's nested `YYYY/MM/DD/` source layout, falling back to mtime if the shape doesn't match.

`github.com/BurntSushi/toml` is promoted from indirect to direct so we can decode existing Codex `notify` values before deciding whether to overwrite them.

## Test plan

- [x] `make fmt vet lint build` clean
- [x] `go test ./internal/sessioncapture/ -v` — unit tests cover: Claude transcript mirror with per-day grouping, Codex mirror of every changed session (not just the global newest), `CODEX_HOME` override, idempotent re-runs that skip up-to-date mirrors, init idempotency, preservation of unrelated Claude keys + hooks, refusal to clobber an existing non-amika Codex notify, in-place update of an amika notify whose path changed, insertion before `[section]` headers, and shell-quoting edge cases.
- [x] End-to-end smoke test from a temp `HOME`: `capture-init` produces the expected `~/.claude/settings.json` Stop hook and `~/.codex/config.toml` notify array; piping a synthetic Claude Stop payload mirrors the JSONL; running the Codex variant mirrors the newest rollout file; re-running `capture-init` is a no-op.
- [ ] Run against a real Claude Code session and verify the Stop hook fires and the mirror lands under `~/.local/state/amika/sessions/claude/<YYYY-MM-DD>/`.
- [ ] Run against a real Codex session and verify the notify program fires and the mirror lands under `~/.local/state/amika/sessions/codex/<YYYY-MM-DD>/`.

## Notes

- Pre-existing `make test` failures in `internal/materialize` come from `rsync` not being installed in this sandbox; verified unrelated by re-running with these changes stashed out.

### DIFF
--- a/go/cmd/amika/sessions.go
+++ b/go/cmd/amika/sessions.go
@@ -1,0 +1,107 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/gofixpoint/amika/go/internal/config"
+	"github.com/gofixpoint/amika/go/internal/sessioncapture"
+	"github.com/spf13/cobra"
+)
+
+var sessionsCmd = &cobra.Command{
+	Use:   "sessions",
+	Short: "Manage local agent session capture",
+	Long: `Capture sessions from local AI coding agents (Claude Code, Codex) into
+the amika state directory by installing per-agent hooks.
+
+Run "amika sessions capture-init" once to register the hooks. From then on,
+each agent will invoke "amika sessions capture" itself whenever it finishes
+a turn; no daemon is involved.`,
+}
+
+var sessionsCaptureInitCmd = &cobra.Command{
+	Use:   "capture-init",
+	Short: "Register session-capture hooks with Claude and Codex",
+	Long: `Install hooks into the local Claude Code and Codex configurations so that
+each agent mirrors its session transcripts into the amika state directory
+whenever it finishes a turn.
+
+Writes to:
+  ~/.claude/settings.json   (adds a Stop hook)
+  ~/.codex/config.toml      (sets the notify program)
+
+This command is idempotent.`,
+	Args: cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return fmt.Errorf("resolving home directory: %w", err)
+		}
+		hookCmd, err := sessioncapture.DefaultHookCommand()
+		if err != nil {
+			return err
+		}
+		rep, err := sessioncapture.Init(home, hookCmd)
+		if err != nil {
+			return err
+		}
+		out := cmd.OutOrStdout()
+		if rep.ClaudeUpdated {
+			fmt.Fprintf(out, "Added Stop hook to %s\n", rep.ClaudeSettingsPath)
+		} else {
+			fmt.Fprintf(out, "Stop hook already present in %s\n", rep.ClaudeSettingsPath)
+		}
+		switch {
+		case rep.CodexConflict != "":
+			fmt.Fprintf(os.Stderr,
+				"Skipped %s: existing notify = %s does not look like amika; leaving it alone\n",
+				rep.CodexConfigPath, rep.CodexConflict)
+		case rep.CodexUpdated:
+			fmt.Fprintf(out, "Set notify program in %s\n", rep.CodexConfigPath)
+		default:
+			fmt.Fprintf(out, "Notify program already set in %s\n", rep.CodexConfigPath)
+		}
+		return nil
+	},
+}
+
+var sessionsCaptureCmd = &cobra.Command{
+	Use:   "capture",
+	Short: "Mirror the current agent session into the amika state directory",
+	Long: `Mirror the session transcript currently being written by Claude Code or
+Codex into the amika state directory.
+
+This is intended to be invoked by the agent's own hook system (see
+"amika sessions capture-init"), not by hand. The --source flag selects
+which agent's session to capture; the Claude variant reads the hook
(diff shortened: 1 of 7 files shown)


jdc123:
Addressed the Codex reviewer's note in dff79bd: `Init` and `CaptureCodex` now route the Codex root through a `CodexHome(homeDir)` helper that honors `$CODEX_HOME` and falls back to `~/.codex`. Added two regression tests via `t.Setenv` covering the override on both the init path (where `config.toml` lands) and the capture path (which `sessions/` tree we scan).

jdc123:
Addressed the latest review note in cebd9c4: added a `useCodexFallback(t)` helper that calls `t.Setenv("CODEX_HOME", "")` and applied it to every test that exercises the `~/.codex` fallback path (`TestCaptureCodex_MirrorsNewestSession`, `TestCaptureCodex_NoSessionsIsNoOp`, and the five `TestInit_*` tests other than `TestInit_HonorsCODEX_HOME`). Verified locally by running the suite with `CODEX_HOME=/opt/codex-poison` set ambiently; all 13 tests pass.

jdc123:
Addressed in b4c5b7b: `Args` on `sessions capture` is now `cobra.MaximumNArgs(1)` so Codex's appended JSON payload (e.g. `'{"type":"agent-turn-complete",...}'`) passes arg validation. The payload is accepted-and-ignored — the mirror still resolves the active session by mtime, which works independent of payload contents. New CLI-level test (`TestSessionsCapture_AcceptsCodexNotifyPayload`) drives the full cobra path end-to-end with a representative payload; a companion negative test ensures >1 positional is still rejected so misconfigurations surface.

jdc123:
Addressed in 4fa5461. Brought the Claude Stop-hook path to parity with the Codex notify path: previously installed amika hooks are now recognized by argv shape (basename `amika` + the `sessions capture --source claude` tail) rather than exact-string equality, so re-running capture-init from a different binary path replaces the stale entry instead of stacking a second one. Empty groups are dropped; duplicate amika hooks collapse to one. Non-amika hooks living in the same Stop group are preserved untouched. New tests cover the path-rename, the non-amika-coexistence, and the duplicate-collapse cases, plus a standalone unit test for the `looksLikeClaudeAmikaHook` matcher across quoted/unquoted/mismatched inputs.

jdc123:
Addressed in 31236c7: `codexNotifyIsAmika` now uses `filepath.Base(argv[0]) == "amika"` instead of `strings.HasSuffix`, matching the basename predicate already used on the Claude side. Lookalikes like `/usr/local/bin/notamika`, `my-amika`, and `amikatool` now route through the existing `CodexConflict` path instead of being silently replaced. Table-driven regression test added (`TestInit_TreatsLookalikeNotifyAsConflict`) covering all three.

jdc123:
Addressed in f93a908. `CaptureCodex` now walks the entire `<codex-home>/sessions/` tree and mirrors every `.jsonl` whose source mtime is strictly newer than its mirror (or that has no mirror yet), instead of picking the single global-newest file. The mirror preserves Codex's `YYYY/MM/DD/<rollout>.jsonl` layout under `<state>/sessions/codex/` so basenames can't collide across days. Unchanged files are skipped via mtime so the steady state remains cheap. Two new tests cover the regression: `TestCaptureCodex_MirrorsAllChangedSessions` (two concurrent sessions on different days both land in one capture) and `TestCaptureCodex_SkipsUpToDateMirrors` (an unmodified session's mirror is not rewritten on subsequent captures, asserted via mtime).

jdc123:
@codex review

dbmikus:
nit: should probably put codex hooks in `~/.codex/hooks.json`

> If more than one hook source exists, Codex loads all matching hooks. Higher-precedence config layers don’t replace lower-precedence hooks. If a single layer contains both hooks.json and inline [hooks], Codex merges them and warns at startup. Prefer one representation per layer.

TBH I wouldn't change this now, but just a heads up that other Codex hooks tools I've used have installed to the `hooks.json`.

For comparison, what hooks does [Entire](https://github.com/entireio/cli) set up?

Do we capture the filepath (or repo name) and the git commit worked?

PR #228, opened 2026-06-04
Open Cursor via a stable per-sandbox SSH alias

dbmikus:
## Summary

Fixes [KAPRO-365](https://linear.app/fixpoint/issue/KAPRO-365/cursor-agent-chat-lost-when-remote-ssh-connection-drops): when a Cursor Remote-SSH connection to an Amika sandbox drops, the agent chat is lost.

**Root cause:** `amika sandbox code` handed Cursor the raw Daytona SSH destination (`ssh-remote+<token>@ssh.app.daytona.io`). Daytona mints a fresh token on every connection, so the destination, and therefore Cursor's remote identity, changed every reconnect. Cursor stores chat client-side keyed to that identity, so it could never re-link an existing session.

**Fix:** give Cursor a stable per-sandbox identity. `amika sandbox code` now writes a managed `Host amika-<sandbox-id>` block to `~/.ssh/amika.conf` and launches Cursor against `ssh-remote+amika-<sandbox-id>`. The alias is keyed on the sandbox's immutable id, so it never changes across reconnects; the rotating Daytona credentials live behind it and are refreshed on each run.

> **Note on the alias key:** I used the sandbox **id**, not the name, because a name can be reused by a future sandbox (which would wrongly inherit a dead sandbox's chat) while the id is immutable and unique.

> **Why not a `ProxyCommand` for auto-refresh:** Daytona puts the rotating token in the SSH *username* (`ssh <token>@ssh.app.daytona.io`), which SSH fixes from config before `ProxyCommand` runs, so a proxy can't refresh it. A stable alias keeps the identity (and chat) stable; an expired token is refreshed by re-running `amika sandbox code`, which rewrites the same block. The companion server PR raises the token lifetime to 8h to make this rare.

## Changes

- **`internal/ssh` package** centralizing how the CLI connects to sandboxes over SSH. `execSSH` moved here as `ExecSSH`; `config.go` adds a JSON-backed host state (`~/.local/state/amika/ssh-hosts.json`, the source of truth), an ssh-destination parser that preserves port/flags, rendering of `~/.ssh/amika.conf` (keyed by id, sandbox name as a comment), and idempotent management of the `Include` line in `~/.ssh/config`. State/config writes are atomic and `0600`.
- **`basedir`** path helpers for the new state file and the `~/.ssh` paths.
- **`apiclient.SSHInfo`** gains `SandboxID`/`SandboxName` (from the server PR). When talking to an older server that omits them, `sandbox code` falls back to a `GetSandbox` lookup so the alias is still id-keyed.
- **`amika sandbox code`** rewired to the stable alias. `amika sandbox ssh` is unchanged (it has no chat-persistence problem).

## Testing

Unit tests alongside each change: destination parsing, config rendering, `Include` injection/idempotency, state round-trip + file permissions, and the full `UpsertHost` flow against temp dirs (no real `~/.ssh`). `go test ./...` passes.

## Manual verification still needed

Needs a live sandbox + Cursor: run `amika sandbox code <name>`, confirm `~/.ssh/amika.conf` + the `Include`, open Cursor, then drop/reconnect Remote-SSH and confirm the agent chat persists.

## Depends on

gofixpoint/amika-mono#454 (server returns `sandbox_id`/`sandbox_name`). Works against an older server via the `GetSandbox` fallback.

### DIFF
--- a/go/cmd/amika/sandbox/sandbox_ssh.go
+++ b/go/cmd/amika/sandbox/sandbox_ssh.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 
+	"github.com/gofixpoint/amika/go/internal/apiclient"
 	"github.com/gofixpoint/amika/go/internal/basedir"
 	"github.com/gofixpoint/amika/go/internal/runmode"
 	"github.com/gofixpoint/amika/go/internal/ssh"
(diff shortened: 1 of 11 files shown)


go/cmd/amika/sandbox/sandbox_ssh.go
```diff
@@ -150,21 +122,45 @@ Examples:
...
+		}
+
+		entry, err := ssh.NewHostEntry(sandboxID, sandboxName, info.SSHDestination, info.ExpiresAt)
+		if err != nil {
+			return err
+		}
+		alias, err := ssh.UpsertHost(basedir.New(""), entry)
+		if err != nil {
+			return fmt.Errorf("write managed SSH config: %w", err)
+		}
```
dbmikus: abstract this out into a function

go/internal/ssh/config_test.go
```diff
@@ -0,0 +1,265 @@
...
+		t.Fatal(err)
+	}
+	if len(state.Hosts) != 1 || state.Hosts[0].User != "token-2" {
+		t.Fatalf("expected single refreshed host, got %+v", state.Hosts)
+	}
+}
+
+// testPaths returns a basedir.Paths rooted in temp dirs so tests touch neither
+// the real ~/.ssh nor the real XDG state directory.
+func testPaths(t *testing.T) basedir.Paths {
```
dbmikus, line 300: Why do we have to set `XDG_STATE_HOME`? Can't we rely on `basedir.New` being correct?

go/internal/ssh/config.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil
+	}
+
+	var content string
+	if len(existing) == 0 {
+		content = includeLine + "\n"
+	} else {
+		content = includeLine + "\n\n" + string(existing)
+	}
+	return writeFileAtomic(configPath, []byte(content), 0o600)
```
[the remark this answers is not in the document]
dbmikus: Fixed in `f1e533a` locally. `EnsureInclude` now resolves a symlinked `~/.ssh/config` to its write target before doing the atomic rename, so the symlink itself is preserved and the managed dotfiles target gets updated. I also added a regression test with a relative symlink to cover the dotfile-manager case.

go/cmd/amika/sandbox/sandbox_ssh.go
dbmikus: Done in `cbdf0cd` locally. I extracted the SSH alias and remote path setup into `prepareCursorSSHTarget`, leaving the Cobra handler focused on validation, launching Cursor, and printing user-facing output.

go/internal/ssh/config_test.go
dbmikus: Good question. `basedir.New(home)` only controls the fallback home-derived paths; it still intentionally honors `XDG_STATE_HOME` from the process environment. If a developer has that env var set, this package test could otherwise write `ssh-hosts.json` under their real state dir. I kept the env override and added a comment explaining that it is there to keep the test hermetic rather than to retest `basedir` itself.

go/cmd/amika/sandbox/sandbox_ssh.go
dbmikus: Done in `cbdf0cd` — extracted into `prepareCursorSSHTarget` which handles the API call, the older-server fallback to `GetSandbox`, and building the `HostEntry`. The command body is now just: call that, call `ssh.UpsertHost`, build the path, exec Cursor.

go/internal/ssh/config_test.go
dbmikus: We need both. `basedir.New(home)` only controls the *fallback* — it makes `StateHome()` return `home/.local/state` when `XDG_STATE_HOME` is unset. But if a developer has `XDG_STATE_HOME` set in their shell environment, `basedir.StateHome()` reads that env var and ignores the home override, so the test would write to their real state directory. Setting `XDG_STATE_HOME` in the test pins the state path to a temp dir regardless of the developer's environment. Added a comment to `testPaths` explaining this.

PR #244, opened 2026-06-12
WIP: amikalog fast push (JSONL sessions, parallel upload, versioning)

dbmikus:
WIP. Adds `specs/006-amikalog-fast-push.md`, the design spec for making `amikalog beta:push` fast and forwards/backwards-compatible. No code changes yet.

## What the spec covers

- **On-disk format:** one append-only `events.jsonl` per session instead of one JSON file per event (kills per-file overhead).
- **Push:** monotone-size sync + idempotent whole-file `upsert`, a small per-session checkpoint replacing the O(N²) per-event manifest, and parallel uploads (batch-signed URLs, bounded worker pool).
- **Versioning:** filename-token container version in the object key + per-line `"v"` record-schema field, with precise additive-vs-breaking rules. v0 ships only the version markers and a "skip unknown, never crash" reader rule; all multi-version handling is deferred until a second format actually exists (nothing is deployed yet).

Crash/partial-line handling, the cross-process flock write model, and v0-vs-future boundaries are documented in the spec.

### DIFF
--- a/specs/006-amikalog-fast-push.md
+++ b/specs/006-amikalog-fast-push.md
@@ -1,0 +1,201 @@
+# amikalog Fast Push: JSONL-per-Session Format, Parallel Upload, and Versioning
+
+## Overview
+
+`amikalog beta:push` is slow because capture writes one file per event and push
+uploads each file serially with two network round-trips (sign + PUT), rewriting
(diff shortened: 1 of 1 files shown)


dbmikus:
Jakub implemented this elsewhere.

PR #248, opened 2026-06-12
amikalog: store one JSONL file per session and parallelize sync

jdc123:
Capture appended one JSON file per hook event, and beta:push/beta:fetch processed objects serially, so a single session produced hundreds of tiny files and hundreds of sequential signed-URL round-trips each way.

Capture now appends one compact JSON line per event to a per-session file (<state>/events/<source>/sessions/<ts>_<session-id>.jsonl) under the existing cross-process lock. Push uploads each changed session file in a bounded parallel pool, tracking each file's uploaded byte size so only sessions that grew new events are re-sent (idempotent upsert); session files are read under the lock so a half-written final line is never uploaded. Fetch downloads each page's objects in a bounded parallel pool, still page-by-page to keep signed URLs fresh.

Backward compatible: legacy per-event event_*.json files are still uploaded, and a tolerant manifest unmarshal maps old timestamp-string entries to "already uploaded" so the format change does not re-push existing history.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -79,7 +79,7 @@ language SDKs live under `sdk/` (e.g. `sdk/typescript/`).
 - `app/` — Application service layer implementation
 - `ports/` — Port interfaces for Docker and store operations
 - `materialize/` — Local sandbox script execution and rsync copying (v0 legacy)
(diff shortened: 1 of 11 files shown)


jdc123:
@codex review

PR #258, opened 2026-06-26
WIP allow syncing of claude memories

dbmikus:


dbmikus:
Note, this PR forked from https://github.com/gofixpoint/amika/pull/255

dbmikus:
@jdc123  I don't think we need to do any `--dangerously-skip-permissions`

You should be able to just do

```
--permission-mode <mode>              Permission mode to use for the session
                                        (choices: "acceptEdits", "auto",
                                        "bypassPermissions", "default",
                                        "dontAsk", "plan")
```

Set to `--permission-mode acceptEdits`

Also, I think Claude has some sandboxing stuff that stop it from editing files outside of the cwd, so depending on that might have to either cd to the directory being modified, or create a tmp directory with both the version A and version B memories and make the merge there. The tmpdir is probably cleaner, so it doesn't leave behind broken stuff

(if we already turned off `--dangerously-skip-permissions`, then my bad for missing it; I was just reading the commit message)

PR #283, opened 2026-07-13
Support remote mode in `service list`

jdc123:
## Problem

`amika service list --sandbox-name <name>` returned **"No services found."** for a running remote sandbox that had a frontend service provisioned (found during the 2026-07-11 release test against staging, `dylan/release-test`).

Root cause: `service list` read **only** the local sandbox state file and never queried the remote API. Remote sandbox creation never writes local state, so the local store is empty for remote sandboxes → the command always reported no services.

Secondary symptom: `sandbox list -l` showed `-` for the PORTS column of remote sandboxes, because the API client (`apiclient.RemoteSandbox`) never parsed the `services` field the server already returns.

## Fix

- **`apiclient`**: parse the sandbox `services` array (`name`, `url`, `hostPort`, `containerPort`, `protocol`) the API already returns — the client was silently dropping it.
- **`service list`**: made remote-first (matching `sandbox list`). Resolves the run mode, gates on auth, and in remote mode fetches each sandbox's provisioned services (with URLs) from the API. Added `--local` / `--remote` / `--remote-target` flags; `--local` preserves the previous local-store behavior.
- **`sandbox list -l`**: populates the remote PORTS column from the sandbox's services instead of `-`. Remote services have no host IP (they are reached via generated URLs), so port bindings omit the misleading `127.0.0.1:` prefix.

## Behavior change

`service list` now defaults to **remote** (consistent with `sandbox list`). Local listing is available via `service list --local`.

## Testing

- `make fmtcheck`, `make vet`, `make lint` clean; unit tests pass.
- Added tests: API JSON-contract parse (`ListSandboxes` services), service-row grouping, port formatting, remote auth gate, and local `service list` behavior.
- Smoke-tested the built binary: local list shows services + URLs, `--sandbox-name` filters, missing sandbox → "No services found.", and default (remote) without credentials fails with `not logged in; run "amika auth login" or use --local`.

Note: the live staging round-trip was not exercisable in the dev environment (no staging credentials), but the remote path is covered by the contract-parse and pure-function tests.


jdc123:
@codex review

PR #335, opened 2026-08-16
Capture container snapshots on the configured target

jdc123:
Container snapshot capture was the one Daytona operation that ran outside the configured region. It went through a client pinned to `target: "experimental"`, because that target alone exposed the capture endpoint when `_experimental_createSnapshot` was the only way to call it. Everything else — sandbox create, snapshot get/list/delete/activate, the VM capture path — already used `DAYTONA_TARGET`.

The SDK graduated the method to `Sandbox.createSnapshot` in 0.202.0 and the endpoint is now served on the regular targets, so the pinned client is gone entirely, along with `createExperimentalDaytonaClient` and `EXPERIMENTAL_DAYTONA_TARGET`. `createDaytonaClient` no longer takes a target override — it has no remaining callers that want one.

## The capture path

Both branches now share one sandbox handle:

```ts
const daytona = createDaytonaClient(config);
const sandbox = await daytona.get(providerSandboxId);
const capture = () => sandbox.createSnapshot(snapshotName, SANDBOX_SNAPSHOT_TIMEOUT_S);
```

Previously the keep-alive branch resolved an experimental handle inside `capture()` while the destructive branch resolved a second, normal handle for the docker quiesce. One handle serves both.

That reorders one thing: the `get` now happens *before* `stopDockerForSnapshot` rather than between the quiesce and the capture. A failure to resolve the handle therefore can't leave a source sandbox with dockerd stopped. The test named for that hazard ("restarts docker even when resolving the experimental handle fails") no longer describes a reachable state, so it's replaced with one asserting docker is never touched at all. The dind quiesce/restore semantics are otherwise unchanged: still destructive-path-only, still restored in `finally`.

## Why this is the same HTTP request

`target` is never a header or a query param in the SDK. It feeds exactly two things: the sandbox-create request body, and `SnapshotService`'s default `regionId`. The capture endpoint is `POST /sandbox/{id}/snapshot`, addressed by id, with `CreateSandboxSnapshot` carrying only `name` and `includeMemory` — no region. So the wire request is byte-identical whichever client sends it. The experimental *client* was never what made capture work; a server-side gate was, and moving the method out of `_experimental_` is the signal that gate is lifted.

## Dependency bump

`@daytonaio/sdk` and `@daytona/api-client` `^0.193.0` → `0.203.0`, the first version carrying stable `createSnapshot` plus a release of settling time.

Exact pins rather than carets: `^0.203.0` resolves to 0.204.1, published two days ago, which amika-mono's `bin/check-new-deps` rejects under its 7-day minimum. This also matches how `prettier`, `vite`, and `vitest` are already pinned in this package.

The `SandboxApi` signatures this package calls directly (`createSandbox`, `getSandbox`, `startSandbox`, `stopSandbox`, `deleteSandbox`, `createSandboxSnapshot`) are unchanged across the bump.

## Verification

`pnpm --filter @amika/sandbox` — 321 tests pass, `typecheck`, `lint`, and `formatcheck` all clean.

## Downstream

amika-mono needs the pointer bump plus its own lockfile update; that PR is stacked on this one and its submodule-pin check stays red until this merges.

## Docstring correction

Review caught that this branch's own capture docstring claimed the snapshot "lands in the configured target". It does not. `this.target` is read in exactly two places in the SDK — `daytona.create()`'s body and `SnapshotService`'s `regionId` default — and neither is on this path. `createSnapshot` addresses the sandbox by id and sends `{name}`, no region. That is the actual reason dropping the pinned client is safe, so the docstring now says that instead.

## Follow-up, split out

The client refactor that came out of this review is now a separate PR stacked on this one, rather than riding along here: **#336**. Under 0.203 the SDK constructor opens a Socket.IO connection, and this package builds a client per operation — worth fixing, but a distinct concern from the region move and separately reviewable.

### DIFF
--- a/js/sandbox/package.json
+++ b/js/sandbox/package.json
@@ -16,8 +16,8 @@
     "test": "vitest run --passWithNoTests"
   },
   "dependencies": {
-    "@daytona/api-client": "^0.193.0",
-    "@daytonaio/sdk": "^0.193.0",
+    "@daytona/api-client": "0.203.0",
+    "@daytonaio/sdk": "0.203.0",
     "@vercel/sandbox": "^2.2.1",
     "freestyle": "^0.1.63",
     "p-retry": "^7.1.1"
--- a/js/sandbox/src/providers/daytona/internal/client.ts
+++ b/js/sandbox/src/providers/daytona/internal/client.ts
@@ -4,32 +4,11 @@
 import { Daytona } from "@daytonaio/sdk";
 import type { DaytonaConfig } from "../config";
 
-/**
- * Target used for the experimental sandbox-snapshot APIs
- * (`sandbox._experimental_createSnapshot`, plus creating new sandboxes
- * off those snapshots). Daytona currently gates these capabilities
- * behind this target regardless of which target the existing image-
- * derived snapshots use.
- */
-export const EXPERIMENTAL_DAYTONA_TARGET = "experimental";
-
-export interface DaytonaClientOptions {
-  /**
-   * Override the configured target for this client. Passed-through value
-   * is honored verbatim — used to switch to the experimental target for
-   * the sandbox-snapshot APIs without rebuilding the entire SDK config.
-   */
-  target?: string;
-}
-
-export function createDaytonaClient(
-  config: DaytonaConfig,
-  options: DaytonaClientOptions = {},
-): Daytona {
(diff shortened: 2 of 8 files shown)


jdc123:
@codex review

jdc123:
@codex review

jdc123:
@codex review

PR #336, opened 2026-08-16
Reuse one Daytona client per config

jdc123:
Stacked on #335 — review that one first; this diff is against it.

## The problem

`@amika/sandbox` builds a fresh `Daytona` on every operation, ~29 construction sites across `operations.ts` and `snapshot-operations.ts`. Under 0.193 that was merely wasteful: the client held nothing but an axios instance.

Under 0.203 the constructor also opens a Socket.IO connection for sandbox state events (`esm/Daytona.js:200-204`, unless `useDeprecatedPolling` is set). 0.193 has no `EventDispatcher` at all — I checked the published package. So on the SDK version #335 brings in, a client per operation means an outbound socket to `app.daytona.io` per Daytona call, each idling out 30s after its listeners drop.

Nothing breaks: the sockets are unref'd, connect failures are swallowed, and state waits still fall back to HTTP polling. But it's connection churn for no benefit, and it's the opposite of what a long-lived client is for.

## The fix

Memoize on the config object's identity with a `WeakMap`:

```ts
const clients = new WeakMap<DaytonaConfig, Daytona>();

export function getDaytonaClient(config: DaytonaConfig): Daytona {
  const cached = clients.get(config);
  if (cached) return cached;
  const client = new Daytona({ ... });
  clients.set(config, client);
  return client;
}
```

Identity rather than field values, for two reasons. The caller already builds one config per process and hands that same object to every operation — amika-mono keeps it on its `ServerEnv` singleton — so identity is the natural key. And it keeps an API key out of a map key. Entries are weak, so a discarded config takes its client with it; there is nothing to close and no lifecycle to manage.

`createDaytonaClient` → `getDaytonaClient`, since it no longer constructs on every call and the old name would be a lie at all 29 sites. `vm.ts` gets the same treatment for its generated `SandboxApi` (`createSandboxApi` → `getSandboxApi`); that one holds no socket, so reuse there only avoids re-allocating, but it keeps both Daytona entry points on one rule.

## Tests

`sandbox-ops.test.ts` needed a change, and the reason is worth knowing: it shared a single config object across tests while re-mocking the SDK constructor per test. The cache defeats that — every test after the first would get the first test's client. Its config is now a factory, which is the honest expression of the new contract. This showed up as a real failure, not a guess.

New `client.test.ts` covers the contract directly: same config → one construction, distinct configs → separate clients, plus target passthrough and the organization-scoping branch.

325 tests pass (up from 321); typecheck, lint, formatcheck clean.

## Note on `useDeprecatedPolling`

Not set, deliberately. One long-lived socket per process is the SDK's intended shape, and the flag is marked `@deprecated` with removal planned. If event streaming ever proves troublesome in practice — filtered egress to the socket.io endpoint would show up as `connect_error` noise and up to 10 reconnect attempts per client — it can be set per-environment via `DAYTONA_USE_DEPRECATED_POLLING` without a code change.

## CI will not run on this PR while it is stacked

`.github/workflows/ci.yml` filters `pull_request` to `branches: [main]`, and this PR targets `amika/fix-regions`. So GitHub reports no checks here — not a failure, just no trigger. CI runs as soon as #335 merges and this retargets to `main`.

Verified locally in the meantime, on this branch alone:

| Check | Result |
| ----------- | ---------------- |
| `test` | 325 passed (35 files) |
| `typecheck` | clean |
| `lint` | clean |
| `formatcheck` | clean |

The downstream amika-mono pointer bump (gofixpoint/amika-mono#967) does get full CI, since that repo's `pull_request` trigger has no branch filter, and it passed there.

## Review follow-ups

A Claude review found no bug and confirmed the parts that actually needed confirming, by reading the SDK source rather than trusting the change:

- **Concurrent sharing is safe.** All `Daytona` instance state is connection-holder state, written once in the constructor and never reassigned. Per-call request options are spread into fresh objects, so nothing mutates shared config. `Sandbox` objects carry nothing back into the client — `get`/`create` hand each one a `structuredClone`d config and its own axios instance, which the SDK does deliberately (there is a comment at `Daytona.js:467` about the constructor mutating `basePath`). The two genuinely shared mutable things, the memoized analytics-URL promise and `EventSubscriptionManager`, are both designed for sharing; the latter is keyed per `resourceId`, so sharing is strictly better here than per-call clients were.
- **Identity is a sound key.** Every consumer path passes the config by reference off a cached env object, and no code anywhere in either repo mutates a `DaytonaConfig` field after construction — so no stale client bound to changed settings.
- **No teardown needed.** `EventDispatcher` unrefs the raw socket and its timers, and falls back to an immediate idle disconnect where it can't. Confirmed empirically: a process that constructs a client and does nothing exits in 38ms.

Three items fixed in response:

1. `providers/freestyle/internal/client.ts` documented its per-call construction as "mirroring the per-call pattern the Daytona provider uses" — no longer true, and it would point a future contributor at the wrong consistency. It now says why Freestyle differs: its client holds no connection.
2. `getDaytonaClient`'s docstring now warns against disposing a cached client. `disconnect()` latches the dispatcher closed permanently, so the cache would keep serving a dead client — and the failure would be quiet, since the subscription manager still issues ids and `waitForState` would wait on a stream that never delivers.
3. `client.test.ts` now asserts the two clients are distinct instances, not merely that the constructor ran twice, and carries the same "factory, not a shared object" warning that `sandbox-ops.test.ts` got.

Informational, no change made: coding-agents ends up with **two** clients, not one — it builds a `WorkerEnv` lazily for Linear events, and `chat-core`'s env calls `sandboxProviderConfigsFromEnv()` again, producing a second config object. Bounded and memoized, but worth knowing the count.

### DIFF
--- a/js/sandbox/src/providers/daytona/internal/client.test.ts
+++ b/js/sandbox/src/providers/daytona/internal/client.test.ts
@@ -1,0 +1,77 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { DaytonaConfig } from "../config";
+import { getDaytonaClient } from "./client";
+
+vi.mock("@daytonaio/sdk", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@daytonaio/sdk")>();
+  return { ...actual, Daytona: vi.fn() };
+});
+
+const { Daytona } = await import("@daytonaio/sdk");
+
+// A factory, not a shared object: the cache is keyed on config identity, so a
+// module-level config would let one case's client bleed into the next.
+const config = (): DaytonaConfig => ({
+  apiKey: "key",
+  apiUrl: "https://daytona.example",
+  target: "us-west-2",
+  organizationId: undefined,
+  useVm: false,
+});
+
+describe("getDaytonaClient", () => {
+  beforeEach(() => {
+    vi.mocked(Daytona).mockReset();
+    // Regular function (not an arrow) so it works as a constructor with `new`.
+    vi.mocked(Daytona).mockImplementation(function () {
+      return {} as InstanceType<typeof Daytona>;
+    });
+  });
+
+  it("builds the client once per config and reuses it", () => {
+    // The point of the cache: the SDK constructor opens a Socket.IO connection,
+    // and operations resolve their client per call. Repeated calls with the
+    // config the caller holds for the process must not each open one.
+    const cfg = config();
+
+    const first = getDaytonaClient(cfg);
+    const second = getDaytonaClient(cfg);
+
+    expect(second).toBe(first);
+    expect(Daytona).toHaveBeenCalledTimes(1);
+  });
+
+  it("builds a separate client per config object", () => {
+    // Keyed on identity, not on field values — two configs that happen to be
+    // field-identical are still two callers, and a caller that rebuilds its
+    // config gets a client matching it.
+    const first = getDaytonaClient(config());
+    const second = getDaytonaClient(config());
+
+    expect(second).not.toBe(first);
+    expect(Daytona).toHaveBeenCalledTimes(2);
+  });
+
+  it("passes the configured target through and omits an unset organization", () => {
+    getDaytonaClient(config());
+
+    expect(Daytona).toHaveBeenCalledWith(
+      expect.objectContaining({
+        apiKey: "key",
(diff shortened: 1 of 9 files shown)


jdc123:
@codex review

jdc123:
@codex review

PR #341, opened 2026-08-17
Run ordinary Daytona commands without a process session

jdc123:
## Why this has to change

A Daytona session owns every process started inside it, and deleting the session
kills them. Every command ran in a throwaway session that was deleted in a
`finally`:

```
create session → run command → delete session (kills its processes)
```

Harmless for `ls`. Fatal for a command whose *entire point* is the daemon it
leaves behind — the provisioning hook that backgrounds OpenCode, `amikad serve
--bg`, the post-snapshot `dockerd` relaunch. Each died the instant its command
returned.

`nohup`/`setsid` don't save it. Those defend against signals and parent death;
Daytona is reaping by ownership, and you can't detach out of a bookkeeping entry.

The result: nothing listening on the agent port, the Coding Agent URL answering
Daytona's `502 proxy upstream error` — and because OpenCode liveness is
warning-only, **provisioning still reported success**. A silent failure.

## The fix

Run ordinary commands through the one-shot `process.executeCommand`. No owning
session, nothing to delete, nothing to kill.

Same command, same sandbox, both paths:

| Path                     | `nohup sleep 600 &`, then probe |
| ------------------------ | ------------------------------- |
| Session (today's `main`) | **dead**                        |
| One-shot (this PR)       | **alive**                       |

## Why it isn't a plain revert

Sessions were adopted (8b995247, 878a08ca) for a real reason: the one-shot API
returns a *single combined* string with stderr folded into the value stream. That
corrupted every caller parsing stdout — `sudo` prints `unable to resolve host` on
images whose hostname doesn't resolve, which broke `amikad host-key show` and took
SSH provisioning down with a 503.

So the two-stream contract is kept and rebuilt on-box: stderr to a temp file, then
stdout + a per-command random marker + that stderr, split by the adapter. The
command sits alone on its own line inside a subshell, so a trailing `&`, a
trailing comment, or a compound command all keep working, and its exit status is
re-raised so the wrapper is invisible. `cwd`/`env` return to being
`executeCommand` parameters.

Both properties now hold at once, at **one round-trip per command instead of
three**.

Two guards had to be carried across explicitly, since neither existed in the
pre-session one-shot code:

- **Env-name validation** moved into `buildDaytonaCommand`. It previously sat only
  in the session builder, which every command reached while they all used
  sessions. Those names are interpolated *unquoted* into `sudo --preserve-env=`,
  so `FOO;touch /tmp/pwned` would end the sudo command and start another. Putting
  it in the builder both paths funnel through is what makes it unbypassable.
- **Stream splitting** is new code (`buildStreamSplitCommand`); the old one-shot
  path had none.

## #342 stays intact

An earlier revision of this PR (`219a37d`) returned non-sudo commands verbatim, on
the reasoning that #342's `bash -c` only guarded bare text in a long-lived session
shell. That was reverted before this PR settled: `buildDaytonaCommand` wraps
**both** paths in `bash -c` at head, byte-identical in behavior to `main`. Nothing
about #342 changes here.

Keeping it also settles four review comments at once — an empty or comment-only
command stays syntactically valid, `kill -TERM $$` reaches the command's own shell
rather than the wrapper, and callers keep guaranteed bash rather than whatever
shell the transport picks.

The relocated reasoning still stands where it matters: the parens in
`executeSessionCommandWithInput` are load-bearing, with a test pinning that shape.
The trigger that motivated #342, `buildRefreshClonedRepoScript`, sends no stdin
and so can no longer reach a session at all.

## What still uses a session, and why

Sending stdin needs the command id, and only an async session run hands that back
before the command finishes. That's the whole remaining set — two credential
writes, both of which consume their bytes and exit rather than daemonizing, so
session teardown costs them nothing:

- `amikad authorized-keys set`
- `amikad connect-token set`

(The pre-existing `streamDaytonaCommandLogs` path also uses a session, for the
same underlying reason: live log streaming needs a command id. Untouched here.)

## Verification

`typecheck`, `lint`, `formatcheck`, 351 unit tests.

Re-verified against a live Daytona sandbox by driving this branch's actual
`executeCommand` — not a reimplementation — on a box created from
`amika-daytona-vm-xs` (2 sandboxes created across two runs, both deleted, 0
leaked):

| Check                                                       | Result                                                                              |
| ----------------------------------------------------------- | ----------------------------------------------------------------------------------- |
| `nohup sleep 601 &`, then a bracketed `pgrep`               | `ALIVE` — the daemon survives                                                       |
| `sudo` on a host whose name doesn't resolve                 | stdout `"OUT\n"`, stderr `"sudo: unable to resolve host sandbox: …\nERR\n"`          |
| `ps -o args= -p "$PPID"` + a real stdout line               | both streams split correctly                                                        |
| `ps aux` + a real stdout line                               | both streams split correctly                                                        |
| `exit 42`                                                   | `exitCode: 42`                                                                      |
| `cwd: /tmp`, `env: {FOO: bar}`                              | stdout `"bar\n/tmp\n"`                                                              |

Row two is the bug that motivated sessions, reproducing live: the real `sudo:
unable to resolve host` warning appears, lands on stderr, and leaves the value
stream clean. That is exactly what `parseEd25519PublicKey` choked on when the two
were merged, so the split is genuinely preserved without sessions.

Not covered: the full provisioning flow end to end — OpenCode listening on 60998,
the Coding Agent URL returning 401 rather than 502, and surviving a stop/start.
This verifies the exec mechanism the fix turns on, not the whole lifecycle above
it.

## Two review findings that don't reproduce

Both were raised against the one-shot path and both were checked on a live
sandbox rather than argued. Neither needs a code change, so nothing was added for
them.

**"Pass a timeout — `ExecuteRequest` defaults an omitted timeout to 10 seconds."**
The type does document that, and the session API it replaces has no timeout field
at all, so the concern is reasonable on paper. The server does not apply it:

| Call                                       | Wall time | Outcome                    |
| ------------------------------------------ | --------- | -------------------------- |
| timeout **omitted**, `sleep 30; echo DONE` | 30450ms   | exit 0, `DONE`             |
| timeout **omitted**, `sleep 45; echo DONE` | 45274ms   | exit 0, `DONE45`           |
| timeout **5**, `sleep 20`                  | 5287ms    | `command execution timeout` |
| timeout **0**, `sleep 30; echo DONE`       | 30286ms   | exit 0, `DONE`             |

The field is live — an explicit `5` kills at 5s — but the documented 10s default
isn't being applied, so no ordinary command is being capped. Nothing on this path
is at risk today, and adding `timeout: 0` would pin behavior that already holds.
Worth knowing if Daytona ever starts honoring its own default.

**"The marker could collide with a command that reports process arguments."**
`splitStreamsAtMarker` cuts at the first occurrence, and the wrapper spells the
marker out in its own script text — so if that text were reachable from inside the
command, `ps aux` would echo the marker and the split would cut in the wrong
place. It isn't reachable: Daytona feeds the script to `zsh` on **stdin**, not as
an argv. Measured with a sentinel placed in the script, from inside the command:

```
parent chain:   pid=995 argv=[/usr/bin/zsh]   pid=891 argv=[/opt/daytona/daytona]
argv_hits:      [NONE]      # grep across /proc/[0-9]*/cmdline
environ_hits:   [NONE]      # grep across /proc/[0-9]*/environ
file_hits:      [NONE]      # grep -r across /tmp /var/tmp /run
```

The `ps aux` and `ps -o args= -p "$PPID"` rows in the table above confirm the
split holds for such commands in practice.

### DIFF
--- a/js/sandbox/src/providers/daytona/internal/adapter.ts
+++ b/js/sandbox/src/providers/daytona/internal/adapter.ts
@@ -32,9 +32,10 @@ export class DaytonaAdapter implements SandboxAdapter {
 
   async exec(command: string, opts?: ExecOptions): Promise<ExecResult> {
     // Delegates to the provider's public exec path so both share one command
-    // builder (sudo → root shell + `--preserve-env`) and one stream-separating
-    // session runner. The SDK's `process.executeCommand` is deliberately not
-    // used: it returns a single combined string, with stderr folded in.
+    // builder (sudo → root shell + `--preserve-env`) and one runner. That
+    // runner calls the SDK's `process.executeCommand`, whose single combined
+    // string it splits back into two streams on-box; see `commands.ts` for why
+    // a process session is the wrong transport for an ordinary command.
     return executeCommand(this.sandbox, command, opts);
   }
 
--- a/js/sandbox/src/providers/daytona/internal/commands.test.ts
+++ b/js/sandbox/src/providers/daytona/internal/commands.test.ts
@@ -1,7 +1,12 @@
 import { describe, it, expect } from "vitest";
+import { spawn, spawnSync } from "node:child_process";
+import { mkdtempSync, readdirSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import {
   buildDaytonaCommand,
   buildDaytonaSessionCommand,
+  buildStreamSplitCommand,
   executeCommand,
 } from "./commands";
 import { fakeDaytonaSandbox } from "./test-support";
@@ -9,25 +14,26 @@ import { fakeDaytonaSandbox } from "./test-support";
 type DaytonaSandbox = Parameters<typeof executeCommand>[0];
 
 describe("buildDaytonaCommand", () => {
-  it("runs a non-sudo command inside its own bash -c", () => {
+  it("wraps a non-sudo command in its own bash -c", () => {
+    // One quoted word, so the command can never interact with the syntax of
+    // whatever script embeds it, and it runs under bash rather than the zsh
+    // Daytona picks for the outer script.
     expect(buildDaytonaCommand("echo hi")).toBe("bash -c 'echo hi'");
     expect(buildDaytonaCommand("echo hi", { sudo: false })).toBe(
       "bash -c 'echo hi'",
     );
   });
 
-  it("confines a command's shell state to a child shell", () => {
-    // A session is a long-lived shell fed commands on stdin, so bare text
-    // would run `set -e` in *that* shell. It then exits before the agent
-    // records an exit status, and the synchronous exec never returns --
-    // `buildRefreshClonedRepoScript` opens with `set -e`, so every
-    // from-snapshot create hung.
-    const cmd = buildDaytonaCommand("set -e\ngit fetch --prune origin");
-    expect(cmd).toBe("bash -c 'set -e\ngit fetch --prune origin'");
+  it("keeps a command that would break an embedding script inert", () => {
+    // Quoted, these cannot swallow a closing paren or strand it after an `&`.
+    expect(buildDaytonaCommand("echo hi # note")).toBe(
+      "bash -c 'echo hi # note'",
+    );
+    expect(buildDaytonaCommand("sleep 1 &")).toBe("bash -c 'sleep 1 &'");
   });
 
-  it("quotes a command containing single quotes", () => {
-    expect(buildDaytonaCommand("git config x 'a b'")).toBe(
+  it("quotes a sudo command containing single quotes", () => {
+    expect(buildDaytonaCommand("git config x 'a b'", { sudo: true })).toContain(
       `bash -c 'git config x '"'"'a b'"'"''`,
     );
   });
@@ -51,6 +57,18 @@ describe("buildDaytonaCommand", () => {
       "--preserve-env=AMIKA_AGENT_CWD,AMIKA_OPENCODE_WEB,OPENCODE_SERVER_PASSWORD,FOO",
     );
   });
+
+  it("rejects an env name that is not shell-legal", () => {
+    // The name lands unquoted in `--preserve-env=`, so a `;` in it would end
+    // the sudo command and start a new one. Checked here, not only in the
+    // session builder, because the no-stdin path never reaches that one.
+    expect(() =>
+      buildDaytonaCommand("true", {
+        sudo: true,
+        env: { "FOO;touch /tmp/pwned": "x" },
+      }),
+    ).toThrow(/Invalid environment variable name/u);
+  });
 });
 
 describe("buildDaytonaSessionCommand", () => {
@@ -68,8 +86,7 @@ describe("buildDaytonaSessionCommand", () => {
     ).toBe(true);
   });
 
-  it("applies cwd and env ahead of the wrapper on the non-sudo path", () => {
-    // The wrapped child inherits both, so they still reach the command.
+  it("applies cwd and env on the non-sudo path", () => {
     expect(
       buildDaytonaSessionCommand("printenv FOO", {
         cwd: "/home/amika",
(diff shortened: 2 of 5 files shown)


jdc123:
## Verified against a live Daytona sandbox

Ran the mechanism directly against a real sandbox (created from
`amika-daytona-vm-xs`, deleted afterwards; 0 leaked), comparing the pre-fix
session path and the one-shot path side by side on the same box:

| Check | Result |
| --- | --- |
| Legacy throwaway session, `nohup sleep … &`, then probe | `alive: false` — **reaped**, confirming the diagnosis |
| One-shot `executeCommand`, same command, then probe | `alive: true` — **survives**, confirming the fix |
| `echo to-stdout; echo to-stderr >&2` | `stdout: "to-stdout\n"`, `stderr: "to-stderr\n"` |
| `sudo` on a host whose name doesn't resolve | `stdout: "ssh-ed25519 AAAA\n"`, `stderr: "sudo: unable to resolve host sandbox: Name or service not known\n"` |
| `exit 42` | `exitCode: 42` |
| `cwd: /tmp`, `env: {FOO: bar}` | `stdout: "bar\n/tmp\n"` |

The fourth row is the original KAPRO-767-adjacent bug reproducing live: the real
`sudo: unable to resolve host` warning appears, and lands on stderr with the value
stream clean. That is exactly what `parseEd25519PublicKey` choked on when the two
were merged, so the split that motivated sessions is genuinely preserved without
them.

The probe deliberately uses a bracketed `pgrep` pattern (`sleep 6[0]1`) so the
asking shell — whose own argv contains the pattern — cannot match itself.

**Still not covered:** the full provisioning flow end to end — OpenCode actually
listening on 60998, the Coding Agent URL returning 401 rather than 502, and
surviving a stop/start. This verifies the exec mechanism the fix turns on, not the
whole lifecycle above it.

jdc123:
## Rebased onto `main`, plus a review fix

Rebased both sides onto current `main`, which had gained the Daytona
session-shell fix (`amika` #342, wrapping the non-sudo path in `bash -c` so a
script's `set -e` can't kill the long-lived session shell). The conflict was
confined to two docstrings; #342's `bash -c` change is kept as-is, and this
change now routes through it.

### A codex review caught a real regression

Moving off sessions meant the no-stdin path stopped going through
`buildDaytonaSessionCommand`, which was the only place environment variable
*names* were validated. Those names are interpolated unquoted into
`sudo --preserve-env=`, so a name like `FOO;touch /tmp/pwned` would have ended
the sudo command and started another one. Every path validated before this
change, so it was a regression, not a pre-existing gap.

Fixed by moving the check into `buildDaytonaCommand` — the builder both paths
funnel through — which is what makes it unbypassable rather than a rule each new
call site has to remember. Covered at both the builder and the `executeCommand`
boundary, and both new tests were confirmed to fail with the guard removed.

### A second finding, checked and not reproduced

The reviewer also flagged that a backgrounded child inheriting stdout would keep
the one-shot request open, calling it a regression against the session path.
Measured on a live sandbox instead of reasoning about it:

| Path | `sleep N &`, no redirect | Blocked? |
| --- | --- | --- |
| One-shot (this PR) | 25429ms for a 25s sleep | yes |
| Session (`main` today) | 26290ms for a 26s sleep | yes |

Both block, near-identically, so the premise that the session path "returned as
soon as the shell command finished" doesn't hold and there is no regression here.
It is a real pre-existing property of Daytona exec worth knowing about — a
command must redirect its background child's stdio to return promptly — but it
is unchanged by this PR and out of scope for it. Every call site that
daemonizes (`opencode-web`, `dockerd`, the agent launches) already redirects
both streams.

### Re-verified live after the rebase

Re-ran the full live check against a fresh Daytona sandbox, since `bash -c`
wrapping changes what actually executes (sandboxes deleted; 0 leaked):

| Check | Result |
| --- | --- |
| Legacy session, `nohup sleep … &` | `alive: false` — reaped |
| One-shot through the `bash -c` wrapper | `alive: true` — survives |
| `echo to-stdout; echo to-stderr >&2` | cleanly split |
| `sudo` on an unresolvable hostname | warning on stderr, value stream clean |
| `set -e` script that fails | `exitCode: 1`, no hang, no unreachable output |
| `cwd` + `env` through `bash -c` | `bar\n/tmp\n` — inherited correctly |
| `exit 42` | `exitCode: 42` |

`typecheck`, `lint`, `formatcheck` and 338 unit tests pass.

jdc123:
@codex review

jdc123:
@codex review

jdc123:
@codex review

PR #349, opened 2026-08-18
Sweep orphaned stderr captures before a snapshot

jdc123:
Follows up the security review of #341, now merged. Fixes the one genuinely new exposure that
change introduces; the framing defects found alongside it are tracked separately
in KAPRO-821.

## The leak

#341 moves ordinary commands onto Daytona's one-shot `process.executeCommand`,
which carries a **single combined** stream. To keep the two-stream contract, the
wrapper sends the command's stderr to a temp file and reads it back once the
command exits.

An `EXIT` trap removes that file, and `INT`/`TERM`/`HUP` are trapped on top. But
`SIGKILL`, an OOM kill, and the sandbox being force-stopped mid-command run **no
trap at all**, and the file stays — with the command's stderr in it. Reproduced
directly:

```
files left in TMPDIR after SIGKILL: [ 'tmp.eBNzYjjt6b' ]
  content: "SECRET_TOKEN_abc123\n"
```

Stderr is not reliably non-sensitive. A failed
`git clone https://x-access-token:<token>@github.com/...` prints the tokenized URL
there — the same class of leak `scrubGitRemoteTokens` already exists to clean out
of `.git/config`. Snapshots capture the filesystem opaquely and are forkable, so
residue under `/tmp` is baked in and travels.

**On the session path #341 replaced, this could not happen**: stderr arrived on
separate log callbacks and never touched the filesystem. #341 introduces
stderr-at-rest, and the core scrub cannot reach it — `ScrubTargets` is a
declarative list of caller-declared credential paths, and nothing sweeps a temp
directory.

## The change

**Name the capture file.** `mktemp` → `mktemp "${TMPDIR:-/tmp}/amika-exec-stderr.XXXXXXXXXX"`.

This is what makes a sweep possible at all. `mktemp`'s default is
`tmp.XXXXXXXXXX`, and the only glob that matches it is `tmp.*` — which would
delete unrelated tools' temp files mid-build. `$TMPDIR` is still honored, so the
existing bail-on-unwritable-temp path and the `cat --` flag guard are unchanged.

**Sweep the leftovers before a capture**, as Daytona's `removeInjectedSecrets` —
the hook `snapshots.scrubAndCreate` runs after the core scrub and immediately
before `capture`. It belongs at the provider layer, not in the shared scrub: that
module removes exactly the paths it is handed and never decides what is a secret,
whereas this path pattern is an invariant of *this* provider's wrapper.

## Why it is not verified fail-closed

The credential scrub verifies removal and fails closed, so the obvious question is
why this doesn't. That scrub removes a known secret at a known path, so a
survivor is a genuine failure. Here every exec — including the verifying one —
creates a capture file of its own while it runs, so a check would be racing its
own plumbing.

That isn't theoretical. Two `ls /tmp` calls through the adapter on a live sandbox,
around the sweep:

```
[before] amika-exec-stderr.0Zpa3c4iZm   <- the listing exec's own capture file
         amika-exec-stderr.orphan01     <- the orphan under test
         important.log
         tmp.someOtherTool

[after]  amika-exec-stderr.SSnC9gLefn   <- a different one, same reason
         important.log
         tmp.someOtherTool
```

A fail-closed check would have thrown on both listings. Each in-flight file is
unlinked by its own trap moments later, so nothing lingers into the capture.

The sweep folds `find`'s stderr into stdout, because the sweep deletes the very
file its own stderr is being captured to; `execFailureText` already prefers stdout
when stderr is empty, so the exit-code check still reports something useful. It
runs unprivileged — the capture file is created by the framing shell before any
`sudo`, so one user owns all of them.

## Verification

`typecheck`, `lint`, `formatcheck`, 359 unit tests.

Unit coverage executes the generated command against a real directory rather than
only asserting the string, and three mutations were confirmed to fail it: globbing
`*` instead of the prefix, dropping the `2>&1` fold, and reverting `mktemp` to
bare. A separate case SIGKILLs a real wrapper and asserts the file it leaves
behind carries the prefix.

Against a live Daytona sandbox (sandbox deleted, 0 leaked):

| Check                                          | Result |
| ---------------------------------------------- | ------ |
| Orphan holding a token removed                 | pass   |
| Unrelated `important.log` survived             | pass   |
| Decoy `tmp.someOtherTool` survived             | pass   |
| Token no longer findable anywhere under `/tmp` | pass   |
| Second sweep on a clean temp dir still succeeds | pass  |

## Known limits

- Bounded to `$TMPDIR` as the sweeping exec sees it. A capture file written under
  a caller-supplied `TMPDIR` is not swept; no caller sets one today, and the
  alternative is searching the filesystem before every snapshot.
- Only the Daytona provider is affected, because only its wrapper writes these
  files. Freestyle and Vercel need nothing.
- Does not address the framing defects from the same review — startup stderr
  landing in stdout, and a background child's stdout landing in stderr. Those
  share a root cause (framing bounded only at the end) and are KAPRO-821.

### DIFF
--- a/js/sandbox/src/providers/daytona/internal/commands.test.ts
+++ b/js/sandbox/src/providers/daytona/internal/commands.test.ts
@@ -4,6 +4,7 @@ import { mkdtempSync, readdirSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
+  STDERR_CAPTURE_PREFIX,
   buildDaytonaCommand,
   buildDaytonaSessionCommand,
   buildStreamSplitCommand,
@@ -285,6 +286,39 @@ describe.skipIf(process.platform !== "linux")(
       expect(stdout).toBe("out\n");
       expect(stderr).toBe("err\n");
     }, 20_000);
+
+    it("names the capture file so the pre-snapshot sweep can find it", () => {
+      // The sweep globs on this prefix. A killed wrapper runs no trap, so the
+      // file survives with the command's stderr in it — and a snapshot taken
+      // afterwards would bake that in. `mktemp`'s default `tmp.XXXXXXXXXX`
+      // could only be swept by a `tmp.*` glob, which would take unrelated
+      // tools' temp files with it.
+      const dir = mkdtempSync(join(tmpdir(), "amika-prefix-"));
+      try {
+        // `sleep` holds the wrapper open; SIGKILL runs no trap, so whatever it
+        // created is still there to inspect.
+        const script = buildStreamSplitCommand(
+          buildDaytonaCommand("echo SENSITIVE >&2; sleep 30"),
+          "--M--",
+        );
+        const child = spawn("sh", { stdio: ["pipe", "ignore", "ignore"] });
+        child.stdin.end(`TMPDIR=${dir}\n${script}`);
+        // Give the wrapper time to reach its `mktemp` before killing it.
+        const deadline = Date.now() + 5000;
+        let names: string[] = [];
+        while (Date.now() < deadline) {
+          names = readdirSync(dir);
+          if (names.length > 0) break;
+          spawnSync("sleep", ["0.05"]);
+        }
+        child.kill("SIGKILL");
+
+        expect(names).toHaveLength(1);
+        expect(names[0]!.startsWith(STDERR_CAPTURE_PREFIX)).toBe(true);
+      } finally {
+        rmSync(dir, { recursive: true, force: true });
+      }
+    });
 
     it("leaves no capture file behind", () => {
       const dir = mkdtempSync(join(tmpdir(), "amika-split-"));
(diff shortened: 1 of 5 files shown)


jdc123:
Superseded by #350.

You were right that this put the decision in the wrong layer. `snapshot-scrub.ts` reserves "what is a secret" for the provisioning layer — core "removes exactly the paths it is handed and never decides which files/env vars are Amika secrets" — and there is a direct precedent I missed: `EXEC_INPUT_STAGING_ROOT` is core's own exec-plumbing residue, exported from `src/index.ts` and declared by `computeSnapshotScrubTargets`. This PR bypassed that entirely.

Following the precedent properly turned out to be blocked anyway: `files` entries are verified fail-closed with `ls -1d`, and *every* exec creates a stderr capture including the verifying one, so the check would trip on its own plumbing and abort every `scrubAndCreate`. Making it work would need a new `ScrubTargets` field with emptiness-rather-than-absence semantics, across both repos.

#350 unlinks the capture file at creation and reads it back through a surviving descriptor instead. There is then no name for an untrappable kill to leave behind, so nothing needs scrubbing, no path constant needs publishing, and no policy decision lands in core. Verified with a real platform-issued kill on a live sandbox.

PR #350, opened 2026-08-18
Unlink the stderr capture instead of trapping its removal

jdc123:
Replaces #349, which fixed the same leak by sweeping the residue before a
snapshot. This removes the residue instead, so there is nothing to sweep.

Follows up the security review of #341.

## The leak

#341 moved ordinary commands onto Daytona's one-shot `process.executeCommand`,
which carries a **single combined** stream. To keep the two-stream contract, the
wrapper sends the command's stderr to a temp file and reads it back once the
command exits.

Removing that file was an `EXIT` trap. `SIGKILL`, an OOM kill, and a sandbox
force-stopped mid-command run **no trap at all**, so the file stayed — with the
command's stderr in it:

```
files left in TMPDIR after SIGKILL: [ 'tmp.eBNzYjjt6b' ]
  content: "SECRET_TOKEN_abc123\n"
```

Stderr is not reliably non-sensitive. A failed
`git clone https://x-access-token:<token>@github.com/...` prints the tokenized URL
there — the same class of leak `scrubGitRemoteTokens` already exists to clean out
of `.git/config`. Snapshots capture the filesystem opaquely and are forkable, so
residue under `/tmp` is baked in and travels.

On the session path #341 replaced this could not happen: stderr arrived on
separate log callbacks and never touched the filesystem.

## The change

Keep a read handle on the capture, then unlink its name immediately:

```sh
__amika_err=$(mktemp) || exit 125
trap 'rm -f -- "$__amika_err"' EXIT        # covers only the window below
exec 2>"$__amika_err" || exit 125          # writes land here
rm -f -- "$__amika_err" || exit 125        # no path resolves to it from here
( command )
… printf marker; cat /proc/self/fd/2       # read the unlinked inode back
```

From the unlink on, the capture exists only as an open inode. The kernel frees it
when the process dies, however it dies, so there is no window in which an
untrappable kill can leave a named file behind. The `EXIT` trap goes away with it;
`INT`/`TERM`/`HUP` stay, since those emit the streams before exiting rather than
clean up.

Three details worth a look, all three shaped by review:

- **Fails closed** if the unlink doesn't take. Proceeding with the name in place
  would reinstate exactly the exposure this prevents.
- **`--` moves from the `cat` to the `rm`**, still guarding a `TMPDIR` beginning
  with `-` from being read as flags.
- **No second descriptor.** An earlier revision kept the read handle on fd 9. That
  needed a free high fd, so `RLIMIT_NOFILE` below 10 broke it — and worse, it was
  inherited by every process the command left running. Since this path exists to
  leave daemons running, a daemon pinned the unlinked inode for its whole life
  *even after redirecting its own streams*, because that replaces fd 1 and 2 and
  never touches fd 9 (`/proc/<pid>/fd/9` still pointed at the deleted capture).
  `/proc/self/fd/2` reopens the same inode with no handle to leak.
- **A trap covers the naming window, and only that.** Between `mktemp` and the
  `rm` the file briefly has a name. A redirection error on `exec` — a special
  built-in — is fatal to a non-interactive shell in `dash` and `sh`, which die
  before any `||` branch can run, so a guard alone cannot clean up there. After
  the `rm` the trap is a no-op on a path that no longer exists, which is why a
  kill past that point still leaves nothing: there is no file, not merely no
  handler.
- **The redirection fails closed.** Bash treats that same failure as non-fatal and
  carries on, and would otherwise run the command and emit its marker with no
  usable capture — a clean success carrying no stderr. Pinned structurally rather
  than executed: redirecting fd 2 onto a descriptor that already exists, over a
  file `mktemp` just created, leaves no reachable failure to drive from a test.

## Why not scrub it, as #349 did

#349 named the capture file and swept the leftovers in Daytona's
`removeInjectedSecrets`. That worked, but it put a "this is sensitive" decision
inside core, which `snapshot-scrub.ts` reserves for the provisioning layer — core
"removes exactly the paths it is handed and never decides which files/env vars are
Amika secrets."

The properly-layered version of that fix would follow `EXEC_INPUT_STAGING_ROOT`:
core exports the path, `computeSnapshotScrubTargets` declares it. But that route
has a real obstacle — `files` entries are verified fail-closed with `ls -1d`, and
*every* exec creates a stderr capture, including the verifying one, so the check
would trip on its own plumbing and abort every `scrubAndCreate`. Making it work
needs a new `ScrubTargets` field with emptiness-rather-than-absence semantics,
across two repos, for residue that only exists after an untrappable kill.

Unlinking up front makes all of that unnecessary: no path constant to publish, no
provisioning change, no new field, and no policy decision in core.

## Verification

`typecheck`, `lint`, `formatcheck`, 356 unit tests.

New unit cases, each confirmed to fail with its own fix reverted: a real wrapper
SIGKILLed with the temp directory asserted empty both *during* the run and after
the kill; a daemon's `/proc/<pid>/fd` checked for any descriptor pinning a deleted
inode; and the wrapper run three times under `ulimit -n 9` **on each of `bash`,
`dash`, `sh` and `zsh`**, asserting zero residue whatever exit status the shell
chooses. That last one is per-shell because the failure is shell-specific and an
earlier version of the test watched only bash — which hid that `dash` and `sh` die
before any `||` branch can run.

Generated script executed under every shell that could run it:

| Shell  | Happy path                                  | Under `ulimit -n 9` (×3 runs) |
| ------ | ------------------------------------------- | ----------------------------- |
| `sh`   | `exit 7` propagates, split, 0 files left    | 0 files left                  |
| `bash` | `exit 7` propagates, split, 0 files left    | 0 files left                  |
| `dash` | `exit 7` propagates, split, 0 files left    | 0 files left                  |
| `zsh`  | `exit 7` propagates, split, 0 files left    | 0 files left                  |

Against a live Daytona sandbox (deleted, 0 leaked). The platform's own command
timeout supplies a genuine untrappable kill, which is the case that matters:

| Check                                                | Result |
| ---------------------------------------------------- | ------ |
| Streams still split correctly                        | pass   |
| Exit status still propagates (`exit 42`)             | pass   |
| `sudo` resolve-host warning stays out of stdout      | pass   |
| No capture file visible even mid-exec                | pass   |
| Wrapper killed untrappably by the platform at ~5s    | pass   |
| Nothing left in `/tmp` after that kill               | pass   |
| The secret its stderr carried is not findable on disk | pass  |
| Daemon inherits no descriptor pinning the unlinked capture | pass |
| Backgrounded daemon still survives the one-shot path  | pass   |
| `cwd` and `env` reach the command                    | pass   |

## Not addressed

A command's own stderr descriptor (fd 2) is still inherited by a background child
that does not redirect, so such a child writes into the unlinked inode and holds
its space until it exits. That is inherent — fd 2 *is* the capture — and it is unchanged from before this PR. It belongs to the
wider framing issue: output written outside the command's synchronous window is
misattributed, because the framing is bounded only at the end. Tracked separately.

### DIFF
--- a/js/sandbox/src/providers/daytona/internal/commands.test.ts
+++ b/js/sandbox/src/providers/daytona/internal/commands.test.ts
@@ -1,6 +1,13 @@
 import { describe, it, expect } from "vitest";
 import { spawn, spawnSync } from "node:child_process";
-import { mkdtempSync, readdirSync, rmSync } from "node:fs";
+import {
+  mkdtempSync,
+  readFileSync,
+  readdirSync,
+  readlinkSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
@@ -159,14 +166,10 @@ describe("buildStreamSplitCommand", () => {
   });
 
   it("cleans up on a signal by exiting, not by falling through", () => {
-    // A POSIX trap handler resumes where the shell was, so a cleanup-only
-    // handler on INT/TERM/HUP would delete the capture file and then run on to
-    // `cat` it — replacing the command's real stderr with a `No such file`
-    // error and reporting its status for a wrapper that was told to stop.
+    // A POSIX trap handler resumes where the shell was, so a handler that fell
+    // through would report the command's own status for a wrapper that was told
+    // to stop.
     const script = buildStreamSplitCommand("true", "M");
-    // The buggy form was `… EXIT INT TERM HUP`, of which this is a prefix, so
-    // pin that EXIT stands alone.
-    expect(script).toContain("trap 'rm -f -- \"$__amika_err\"' EXIT\n");
     for (const [signal, code] of [
       ["INT", 130],
       ["TERM", 143],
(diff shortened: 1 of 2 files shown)


jdc123:
@codex review

PR #355, opened 2026-08-19
Restore E2B image permissions

dbmikus:
## Summary
- restore the E2B sandbox image filesystem contract during creation
- clean up a newly created sandbox if restoration fails
- cover successful restoration and failure cleanup

## Validation
- pnpm --filter @amika/sandbox test
- E2B API CLI E2E suite (supported cases)


dbmikus:
Closing because the permission repair now lives in amika-mono sandbox-provisioning, where Amika-specific lifecycle behavior belongs. The parent branch restores this submodule to the existing main commit.

PR #361, opened 2026-08-20
add vs code editor support in amika cli

spline2hg:
adds --editor vscode to sandbox code and codev2


dbmikus:
Looks good to me! We're merging and will include in the next CLI release today (v0.15.0)

PR #373, opened 2026-08-20
Restore Pi Web's request identity behind the proxy

jdc123:
## The bug

Selecting a project in the Pi Web UI answers `403 {"error":"Untrusted API request"}` on every sandbox reached through Daytona's proxy. `POST /api/cwd/validate` fails; the page itself loads fine.

Pi Web guards `/api/*` by requiring the browser's `Origin` to equal `<scheme from X-Forwarded-Proto>://<Host header>` (`isApiRequestOriginAllowed`, Pi Web's `lib/request-security.ts`). Both halves are whatever the provider's proxy makes them, and a header capture — taken by standing a recording server on the mapped port — shows Daytona keeps neither:

```
host: localhost:60996                                     <- external host erased
origin: https://60996-<sandbox>.daytonaproxy01.net        <- real browser origin
x-forwarded-port: 443    (no x-forwarded-proto, no x-forwarded-host)
```

Pi Web therefore expects `http://localhost:60996` and compares it against an `https://…daytonaproxy01.net` Origin: mismatched on host *and* scheme. Browsers omit `Origin` on same-origin GETs, so reads pass and every write 403s, which is exactly the reported symptom.

`AMIKA_PI_WEB_ALLOWED_HOSTS` cannot help. It feeds only Pi Web's separate Host check, and that check already passes on the rewritten loopback Host — so the allow-list was not what made the endpoint work.

## Why a shim

Pi Web has no origin allow-list to configure instead: `lib/request-security.ts` reads only `PI_WEB_HOSTNAME` and `PI_WEB_ALLOWED_HOSTS`, and 0.8.9 (what `versions.env` pins) is the latest release. Upstream [pi-web#497](https://github.com/agegr/pi-web/issues/497) is the same defect, open since 14 Aug with no maintainer response; [#511](https://github.com/agegr/pi-web/pull/511) would not fix our variant (it requires `X-Forwarded-Proto` to be present) and [#544](https://github.com/agegr/pi-web/pull/544) would, but both are unreviewed.

## The change

Pi Web moves to **60995 on loopback**; a shim owns the mapped port **60996** and restores the `Host` and scheme the browser actually used before Pi Web sees the request.

The rewrite only ever targets a host already trusted: the public hostname Amika minted for this sandbox (`AMIKA_PI_WEB_ALLOWED_HOSTS`), or the `Host` the request already carries. An `Origin` naming anything else is forwarded untouched and still rejected by Pi Web, so a cross-origin page gains nothing by asking, and Pi Web's Host allow-list — where the CSRF and DNS-rebinding signal actually lives — keeps doing its job on a `Host` that now means something. An existing `X-Forwarded-Proto` is left alone.

The shim also tunnels `Upgrade` connections and pipes without buffering so streaming responses stay live, and `pi-setup.sh` takes both halves down together so a failure closes the port instead of serving 502s.

## Test plan

Verified against a live Daytona sandbox through the public HTTPS URL (not loopback — the bug is invisible over loopback, since it lives in the headers the proxy rewrites):

| Case | Before | After |
| --- | --- | --- |
| `POST /api/cwd/validate`, browser-like `Origin` | 403 `Untrusted API request` | **200** `{"success":true,…}` |
| Same, unauthenticated | 403 | 401 (trust gate passes, auth still enforced) |
| `GET /` page load | 200 | 200 |
| `POST` with `Origin: https://evil.example` | 403 | **403** |
| `POST` with `Sec-Fetch-Site: cross-site` | 403 | **403** |
| `POST` with lookalike `…daytonaproxy01.net.evil.example` | 403 | **403** |
| API reads (`/api/sessions`, `/api/home`) | 200 | 200 |
| SSE `/api/agent/running/events` | — | events delivered, `text/event-stream` preserved |

Also run: `python3 sandbox-image/generate.py --check` (generated files current), `make test-sandbox-image` (bundle consistent, verification suite passes), `make shellcheck` (clean), `node --check` on the shim.

`make vet`/`lint`/`build` were not run — no Go toolchain in the sandbox this was developed in, and no `.go` sources changed (the `go/internal/sandbox/sandbox-image/` files are the generated asset mirror). CI covers them.

## Notes

- No control-plane change needed: the mapped port and service URL are unchanged, so `PI_WEB_PORT` in amika-mono's `sandbox-provisioning` still points at the right place. Its doc comment there could gain a line about 60995 being taken.
- The shim can be deleted once an upstream release compares hosts rather than full origins; the docs say so at the call site.
- Not addressed: [pi-web#542](https://github.com/agegr/pi-web/issues/542), where Chromium 150+ strips the port from `Origin`. It would bite local-docker sandboxes on a non-default port, and it is a separate upstream bug — the shim deliberately compares authority including port rather than papering over it.

### DIFF
--- a/docs/sandbox-configuration.md
+++ b/docs/sandbox-configuration.md
@@ -213,13 +213,14 @@ Amika reserves container ports **60899–60999** (101 ports) for internal
 services that run inside sandboxes. User workloads and setup scripts should
 avoid binding to ports in this range.
 
-| Port        | Service                          | Status   |
-| ----------- | -------------------------------- | -------- |
-| 60999       | amikad daemon                    | Reserved |
-| 60998       | OpenCode web UI                  | Active   |
-| 60997       | amikad managed sshd (loopback)   | Active   |
-| 60996       | Pi Web UI                        | Active   |
-| 60899–60995 | _(unassigned, reserved for use)_ | Reserved |
+| Port        | Service                            | Status   |
+| ----------- | ---------------------------------- | -------- |
+| 60999       | amikad daemon                      | Reserved |
+| 60998       | OpenCode web UI                    | Active   |
+| 60997       | amikad managed sshd (loopback)     | Active   |
+| 60996       | Pi Web UI (X-Forwarded-Proto shim) | Active   |
+| 60995       | Pi Web itself (loopback)           | Active   |
+| 60899–60994 | _(unassigned, reserved for use)_   | Reserved |
 
 The OpenCode web server starts automatically on port 60998 when OpenCode is
 installed in the container and `AMIKA_OPENCODE_WEB` is not set to `0`. The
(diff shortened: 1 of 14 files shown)


jdc123:
Superseded by gofixpoint/amika-mono#1029: the shim is now injected from the control plane at provision time (the deployed base snapshots are fixed, so an image change can't reach existing fleets). The image-baked arrangement in this PR remains the eventual shape once a base-snapshot rebuild ships the stack; the injected implementation notes that fold-in explicitly.

PR #388, opened 2026-09-01
Give the TypeScript SDK parity with the amika CLI

jdc123:
## Why

`@amika/sdk` advertises itself as a 1:1 port of `go/internal/apiclient`, but it covered roughly two thirds of that client. Several `amika` commands had no TypeScript equivalent at all, and the shapes the SDK *did* return silently dropped fields the API sends — a caller could not see a sandbox's services, its mounted secrets, or a snapshot's id.

## What's new

| Area | Methods | CLI commands they back |
| ---- | ------- | ---------------------- |
| Repositories | `listRepositories` | `amika snapshot list` (repo resolution) |
| Snapshots | `getSandboxSnapshot`, `waitForSandboxSnapshot` | `amika snapshot create --wait` |
| Services | `listSandboxServices`, `createSandboxService`, `putSandboxService`, `deleteSandboxService` | `amika service create\|list\|delete` |
| SSH keys | `createSSHPublicKey`, `listSSHPublicKeys`, `deleteSSHPublicKey` | `amika secret ssh-key push\|list\|delete` |
| SSH dial | `createSSHSession` | `amika ssh`, `amika code`, `amika scp` |
| Agent sessions | `sendAgentSession`, `sendAgentSessionStream`, `listAgentSessions`, `getAgentSession` | `amika send`, `amika sessions list\|show` |

Two supporting modules come with them, neither adding a runtime dependency:

- `src/sse.ts` — SSE frame parsing for the streaming send, with the same rules as Go's `readAgentSessionStream` (blank-line framing, one optional space after the colon, CRLF tolerance, event-less frames dropped without bleeding into the next).
- `src/ssh-session.ts` — descriptor validation plus `canonicalEd25519PublicKey`, an OpenSSH ed25519 parser written against the wire format rather than pulling in a crypto library.

## Schema fidelity

`RemoteSandbox` gained the ~20 fields it was dropping (`services`, `mountedSecrets`, `setupStatus`, the GitHub auth columns, and the rest); `SandboxSnapshot` gained `id`, `repositoryUrl`, `captureMode`, `daytona`; `Secret` is now the full `SecretSummary`; `SandboxScrubPreview` decodes `restoredFiles`; `AgentSendResponse` decodes `isNewSession`, `agentSessionId`, `costUsd`; `CreateSandboxRequest` accepts `githubAuthMode`.

## Breaking change

Nullability now follows the API schema instead of being flattened. A schema-nullable field is `T | null`; an optional one is `T | undefined`. Readers of `repoUrl`, `provider`, `branch`, `commitHash`, and the other nullable columns previously got `""` where the API said `null` and must now handle `null`. The README documents the rule.

No version bump here — releases land as their own commit via `create-sdk-release-commit`.

## Out of scope

`amikalog`'s storage endpoints (a separately installed CLI), the `auth login` OAuth device flow, and the local Docker commands (`materialize`, `volume`, local `sandbox`). None of those are remote API calls an HTTP client can make.

## Testing

`pnpm run ci` passes: 120 unit tests (up from 52), typecheck, lint, formatcheck, build. New coverage for every added method and mapper, the SSE parser, and the ed25519/descriptor validation.

Functional tests were **not** run against a server. `test/functional/org-resources.functional.test.ts` is new and covers the read-only listings plus an SSH-key round trip; it provisions nothing, so it is cheap to run on staging.

### DIFF
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -59,45 +59,116 @@ new AmikaClient({
 
 ## API surface
 
-Methods on `AmikaClient` mirror Go's `*apiclient.Client` 1:1:
+Methods on `AmikaClient` mirror Go's `*apiclient.Client` 1:1, so everything the `amika` CLI can do over the network is reachable from here.
 
-| Method                                | Endpoint                                              |
-| ------------------------------------- | ----------------------------------------------------- |
-| `listSandboxes()`                     | `GET /sandboxes`                                      |
-| `createSandbox(req)`                  | `POST /sandboxes`                                     |
-| `getSandbox(name)`                    | `GET /sandboxes/{name}`                               |
-| `waitForSandbox(name)`                | polls `GET /sandboxes/{name}` until ready             |
-| `getSSH(name)`                        | `POST /sandboxes/{name}/ssh`                          |
-| `revokeSSH(name, token)`              | `DELETE /sandboxes/{name}/ssh`                        |
-| `startSandbox(name)`                  | `POST /sandboxes/{name}/start`                        |
-| `waitForSandboxStart(name)`           | polls until ready                                     |
-| `stopSandbox(name)`                   | `POST /sandboxes/{name}/stop`                         |
-| `waitForSandboxStop(name)`            | polls until `stopped`                                 |
-| `deleteSandbox(name)`                 | `DELETE /sandboxes/{name}`                            |
-| `listSecrets()`                       | `GET /secrets`                                        |
-| `createSecret(req)`                   | `POST /secrets`                                       |
-| `updateSecret(id, req)`               | `PUT /secrets/{id}`                                   |
-| `createProviderSecret(provider, req)` | `POST /secrets/{provider}`                            |
-| `listProviderSecrets(provider)`       | `GET /secrets/{provider}`                             |
-| `deleteProviderSecret(provider, id)`  | `DELETE /secrets/{provider}/{id}`                     |
-| `agentSend(name, req)`                | `POST /sandboxes/{name}/agent-send` (10-min timeout)  |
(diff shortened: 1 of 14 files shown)


jdc123:
### Heads-up: the `SDK TypeScript` check fails for a reason unrelated to this diff

```
Error: Multiple versions of pnpm specified:
  - version 11 in the GitHub Action config with the key "version"
  - version pnpm@10.18.2 in the package.json with the key "packageManager"
```

`pnpm/action-setup` reads the **repo-root** `package.json`, not the job's `working-directory`. Once the root gained a `packageManager` pin (in the `@amika/sandbox` import), `.github/workflows/sdk-typescript.yml` had two versions in play and the action refuses to run. `sdk-typescript-publish.yml` carries the same `version: 11` and will fail the next release the same way.

Nothing under `sdk/typescript/` had changed since the pin landed, so this PR is simply the first run to surface it.

The one-line fix is to drop the `version` input and point the installer at the SDK's own pin, matching what root `ci.yml` already does:

```yaml
      - uses: pnpm/action-setup@v4
        with:
          package_json_file: sdk/typescript/package.json
```

I could not include it here — pushing workflow changes needs the `workflows` permission, which this token lacks. It wants to land either as a separate PR or as a commit pushed onto this branch by someone with that permission.

In the meantime the same pipeline passes locally against `sdk/typescript/pnpm-lock.yaml`:

```
$ pnpm run ci
✓ typecheck  ✓ lint  ✓ formatcheck  ✓ build
Test Files  6 passed (6)
     Tests  120 passed (120)
```

jdc123:
Closing: this is far too large to review in one pass. Reopening as a stack of small PRs, one per CLI command surface (sandbox, secret, agent-send, snapshot, service, secret ssh-key, ssh/code/scp, send/sessions).

PR #395, opened 2026-09-01
Let the SDK open an SSH dial into a sandbox

jdc123:
**Stack 7/8** — based on #394.

## Why

`amika ssh`, `amika code`, and `amika scp` all begin by asking the API for a one-shot transport descriptor, then hand it to OpenSSH or a WebSocket dialer. Without `createSSHSession` a TypeScript caller could provision a sandbox but not reach into it.

## What

`createSSHSession(sandboxId)`, plus `src/ssh-session.ts`.

The descriptor is validated before it is returned, because acting on a bad one is exactly what the check guards against: a wrong `sandboxId` would dial someone else's sandbox, and a malformed connect URL, credential, or host key would weaken the dial. The rules mirror Go's `SSHSession.Validate` field for field.

Validating the host key means parsing an OpenSSH ed25519 line. `canonicalEd25519PublicKey` does that against the SSH wire format rather than by pulling in a crypto library, so the SDK still has zero runtime dependencies. It is exported too — run a key through it before `createSSHPublicKey` and a typo fails locally, in the same canonical comment-stripped form the server stores.

## Testing

`pnpm run ci` passes locally (98 unit tests). 20 of them are new: the ed25519 parser against a real key, an RSA key, authorized_keys options, a relabelled blob, and a table of 11 rejection cases for the descriptor validator.

---

### The stack

Merge front to back. Each PR targets the one above it; GitHub retargets automatically as they land.

| # | PR | Command surface |
| - | -- | --------------- |
| 1 | #389 Decode every sandbox field the API returns | `amika sandbox list\|create\|get` |
| 2 | #390 Return the whole secret summary, not three fields of it | `amika secret list\|push` |
| 3 | #391 Report what an agent-send turn actually cost | `amika sandbox agent-send` |
| 4 | #392 Let the SDK follow a snapshot capture to completion | `amika snapshot` |
| 5 | #393 Expose sandbox services to the TypeScript SDK | `amika service` |
| 6 | #394 Manage SSH public keys from the TypeScript SDK | `amika secret ssh-key` |
| 7 | #395 Let the SDK open an SSH dial into a sandbox | `amika ssh` / `code` / `scp` |
| 8 | #396 Add the agent-session chat surface to the SDK | `amika send`, `amika sessions` |

Together they close the gap between `@amika/sdk` and the network surface of the `amika` CLI. Deliberately out of scope: `amikalog`'s storage endpoints (a separately installed CLI), the `auth login` OAuth device flow, and the local Docker commands (`materialize`, `volume`, local `sandbox`) — none are remote API calls an HTTP client can make.

No version bump in any of these; releases land as their own commit via `create-sdk-release-commit`.

> **On the missing checks:** both `ci.yml` and `sdk-typescript.yml` trigger only on `pull_request: branches: [main]`, so PRs 2–8 show no checks while they target their parent branch. Each picks up CI when GitHub retargets it to `main` as its parent merges.

### DIFF
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -91,6 +91,7 @@ Methods on `AmikaClient` mirror Go's `*apiclient.Client` 1:1:
 | ------------------------- | -------------------------------------- |
 | `getSSH(name)`            | `POST /sandboxes/{name}/ssh`           |
 | `revokeSSH(name, token)`  | `DELETE /sandboxes/{name}/ssh`         |
+| `createSSHSession(id)`    | `POST /sandboxes/{id}/ssh-sessions`    |
 | `createSSHPublicKey(req)` | `POST /secrets/ssh-public-keys`        |
 | `listSSHPublicKeys()`     | `GET /secrets/ssh-public-keys`         |
 | `deleteSSHPublicKey(id)`  | `DELETE /secrets/ssh-public-keys/{id}` |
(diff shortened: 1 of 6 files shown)


jdc123:
Closing: the SSH work is punted to the backlog.

The blocker is that this stack only ports half of the v2 SSH flow. `createSSHSession` (#395) returns and validates the transport descriptor, but connecting to a sandbox also needs a WebSocket dialer, host-key pinning, and a stdio bridge, which live in `go/internal/ssh` (~440 lines in `session.go` plus the dialer) and have no TypeScript equivalent. Without that, `createSSHSession` gets an SDK caller a ticket but not a ride, and the legacy `getSSH` / `revokeSSH` routes stay the only working path.

Rather than land a partial v2 surface that cannot yet replace v1, the whole area is deferred. Tracked in the Linear ticket linked below, which covers the dialer port, the SSH public key endpoints from this PR, and retiring the v1 routes once a real replacement exists.

The branch stays on the remote, so nothing here is lost. #396 has been rebased onto `main` and no longer depends on this.

jdc123:
Tracked in [KAPRO-882](https://linear.app/fixpoint/issue/KAPRO-882/typescript-sdk-port-the-v2-ssh-dialer-then-retire-the-v1-ssh-routes).

PR #400, opened 2026-09-01
Drop SSH support from the TypeScript SDK

jdc123:
Removes the SDK's SSH surface entirely. We will add it back once the v2 dialer is ported — tracked in [KAPRO-882](https://linear.app/fixpoint/issue/KAPRO-882/typescript-sdk-port-the-v2-ssh-dialer-then-retire-the-v1-ssh-routes).

## Why

The only SSH the SDK exposed was **v1**: `getSSH` / `revokeSSH`, backed by `POST` and `DELETE /api/v0beta1/sandboxes/{id}/ssh`. In the CLI those routes are reached only by `sshv1`, `codev1`, and `scpv1` — all `Hidden: true` since `c7fee84` promoted the v2 commands to `ssh`, `code`, and `scp`. So a published SDK was pointing new code at the legacy path.

v2 can't take over yet. `createSSHSession` gets you a transport descriptor, but connecting also needs a WebSocket dialer, host-key pinning, and a stdio bridge — ~440 lines in `go/internal/ssh/session.go` plus the dialer, with no TypeScript equivalent. Shipping half of v2 alongside all of v1 is worse than shipping neither, so this removes SSH and defers the whole area.

## What goes

| Removed | |
| --- | --- |
| `getSSH(name)` | `POST /sandboxes/{name}/ssh` |
| `revokeSSH(name, token)` | `DELETE /sandboxes/{name}/ssh` |
| `SSHInfo`, `RevokeSSHRequest`, `sshInfoFromWire` | types + mapper |

**Breaking** for anyone calling those, with no in-SDK migration path until the dialer lands. The server still serves both routes, so a caller who needs them can hit the API directly in the meantime.

## Coverage this costs

`release.functional.test.ts` used SSH to write a sentinel file into a sandbox, capture a snapshot, boot a new sandbox from it, and read the file back. That was **the only check that a capture actually preserves filesystem contents**, and it goes with the SSH removal.

What survives: the snapshot still gets captured, polled to `active`, and booted from, and that boot is still asserted. What is no longer verified is what is *inside* the snapshot.

If that matters before the dialer lands, `agentSend` could write and read the sentinel instead — slower, billable, and less deterministic since an agent interprets the instruction, but it would restore the assertion. Say the word and I will add it; I did not want to swap a precise check for a fuzzy one without asking.

## Testing

`pnpm run ci` passes: typecheck, lint, formatcheck, build, 82 unit tests (down from 84 — the two SSH tests are gone). I also typechecked the functional tier separately, since `pnpm test` excludes it, and confirmed the built `dist/` exports no SSH symbols (32 client methods, none matching `/ssh/i`).

## Note on #396

Based on `main`, parallel to [#396](https://github.com/gofixpoint/amika/pull/396). Whichever merges second may need a trivial rebase — both touch the README API tables and `src/index.ts`'s export block, in different regions.


jdc123:
Rebased onto `main` now that #396 has merged. The two conflicted where I expected, both in `README.md`, and both are resolved by keeping main's post-#396 text minus the SSH parts.

One extra fix while in there. Main currently claims:

> `org-resources.functional.test.ts` is the exception: it only reads org-scoped listings **and round-trips one SSH public key** …

That is no longer true. The SSH key round trip lived in #394, which was closed rather than merged, so the sentence describes a test that does not exist. Corrected here, since this is the SSH PR.

Re-verified after the rebase: `pnpm run ci` green (119 unit tests), the functional tier typechecks separately, and the built `dist/` exports 36 client methods with none matching `/ssh/i`. `grep -i ssh` over `src/`, `test/` and the README returns nothing.

PR #402, opened 2026-09-02
Infer the git repo for `amika send`, and give it --git/--no-git

jdc123:
`amika send "What does this repo do?"` from inside a git repo created a
sandbox with no repo in it, so the agent had nothing to look at. It now
resolves a repo through `internal/gitrepo`, the same path `amika sandbox
create` takes, and accepts the same --git / --no-git pair.

Closes KAPRO-879, KAPRO-880.

## What `send` does now

- Run it inside a repo and the repo's `origin` is sent as `repo_url`, so
  the sandbox the API creates clones it. Detection walks up from the cwd,
  as create's does.
- `--git <path|url>` picks a different repo and `--no-git` sends no repo
  at all.
- `--repo` becomes a hidden alias for `--git`, so existing callers keep
  working. Passing both with different values is an error.
- Repo selection only applies to a sandbox this command creates, so with
  `--session-id` or `--sandbox` auto-detection is skipped rather than
  silently discarded, and an explicit `--git`/`--repo` is rejected with a
  message saying why. `--no-git` is accepted there — it asks for exactly
  what already happens.
- Resolution runs before the auth gate (matching create) so a
  contradictory flag pair fails on its own terms instead of behind a
  login error; the git subprocess that resolves the URL runs after it.
- In text mode the repo is named on stderr as `repo: <name>`, next to
  `session_id:`; `-o json` suppresses it so stdout stays one valid
  object. The name rather than the URL, matching what `sandbox create`
  reports — it is what identifies the repo to a reader, and a URL would
  risk echoing a token, since an origin can carry one in its userinfo.

`AddRepoAlias` and `FlagsChanged` join `internal/gitrepo` here rather
than in the extraction commit, since the alias and the
"was a repo actually requested?" question exist only for this command.

### DIFF
--- a/go/cmd/amika/send.go
+++ b/go/cmd/amika/send.go
@@ -13,6 +13,7 @@ import (
 	"text/tabwriter"
 
 	"github.com/gofixpoint/amika/go/internal/apiclient"
+	"github.com/gofixpoint/amika/go/internal/gitrepo"
 	"github.com/gofixpoint/amika/go/internal/output"
 	"github.com/gofixpoint/amika/go/internal/runmode"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


jdc123:
Closing in favor of three smaller PRs, so each change is reviewable on its own:

1. Auto-infer the repo from the cwd, plus `--no-git` (the opt-out for the new default) — KAPRO-879
2. `--git <path|url>` for explicit selection — KAPRO-880
3. `--repo` as a hidden alias for `--git`

The two `internal/gitrepo` prerequisites from this branch already landed separately as #404 (extraction) and #405 (the scp-URL classifier bug).

PR #407, opened 2026-09-02
Warn when `amika send` will not have the branch you are on

jdc123:
`amika send` clones the repo's default branch, because the agent-sessions
API accepts no branch. Asking about work in progress therefore gets an
answer about different code, and the only clue was the repo name, which
looks exactly right. This cannot be fixed in the CLI, so it is at least
made visible:

    $ amika send "why does the parser drop escapes?"
    repo: amika
    warning: you are on "fix-parser" but the sandbox will get "main" — the
             agent-sessions API cannot select a branch yet.
             Push "fix-parser" and mention it in your message, or use
             "amika sandbox create --branch fix-parser" instead.

## Staying quiet when unsure

A warning shown to someone already on the default branch is pure noise,
so the comparison is only made when it can be made confidently. Nothing
is printed for a `--git <url>` source (no local checkout), a detached
HEAD (no branch name), or a repo whose origin records no default branch.

`gitrepo.DefaultBranch` reads the local `refs/remotes/origin/HEAD` symref
that git writes at clone time, so it costs no network round trip on a
command that would pay it every invocation. A `git init` repo with a
manually added remote has no such symref and returns an error rather than
guessing at "main" — guessing would put a wrong branch name in front of
someone whose default is "develop".

Suppressed under `--output json` like the other progress output.

### DIFF
--- a/go/cmd/amika/send.go
+++ b/go/cmd/amika/send.go
@@ -54,9 +54,11 @@ from --agent, else the organization's default, else claude.
 Run from inside a git repo and that repo's "origin" remote is cloned into
 the sandbox, the same way "amika sandbox create" picks up the repo you are
 standing in. Pass --git <path|url> to choose a different repo, or --no-git for
-a sandbox with no repo at all; --repo is accepted as an alias for --git. Only
-the repo's default branch is cloned, whatever branch you have checked out
-locally.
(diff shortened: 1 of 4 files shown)


jdc123:
Superseded: the agent-sessions API now takes a branch, so `amika send` can select one rather than only warning that it cannot. Replaced by the branch-support PR.

PR #426, opened 2026-09-07
Remove the container when create fails to save sandbox state

### LINKED ISSUE #425: Failed sandbox create leaves an orphan container and volume refs on Store.Save failure
## Problem

When the sandbox record write fails at the end of a create, both create paths leave the Docker container behind with no record pointing at it:

1. CLI: go/cmd/amika/sandbox/sandbox_create.go returns `sandbox created but failed to save state` right after `store.Save` fails, without removing the container or rolling back volume refs. `rollbackVolumes()` exists in the same function and is called on every earlier failure, and `rb.Disarm()` sits just below the error return, but neither runs on this path.
2. Service/API: `serviceImpl.CreateSandbox` (go/pkg/amika/service.go) returns `ErrInternal` on `s.sandboxes.Save` failure, skipping even the setup-script and git-clone temp dir cleanups it runs on every earlier failure.

The result is an orphan: retrying the create fails with a docker name collision, and there is no amika command that can reach the leftover container or its `amika-git-*` / `amika-rwcopy-*` volume refs.

## Repro

1. Make the state directory unwritable.
2. Run `amika sandbox create --local --name demo --image ubuntu:latest --yes`.
3. The create fails with the save-state error, but `docker ps -a --filter name=demo` still shows the container.

## Expected

On the save failure, the create must roll back what it built (volume refs, rwcopy mounts, temp dirs) and remove the container before returning, keeping the original save error in the message. If removing the container itself fails, the error should say so explicitly so the user knows to clean up manually.

## Suggested fix

- CLI: call the existing `rollbackVolumes()` and `sandbox.RemoveDockerSandbox(name)` on the `store.Save` error path.
- Service: call `cleanupSetupScript()`, `cleanupGitRepo()`, and `sandbox.RemoveDockerSandbox(name)` on the `s.sandboxes.Save` error path.
- A Docker-gated integration test can force the failure via a read-only `AMIKA_STATE_DIRECTORY` (running from a non-git working directory keeps the create bare so the record write is the only store write) and assert no container remains under the requested name.

We have a patch and can share it on Discord.

warun7:
## Problem

When the sandbox record write fails at the end of a create, both create paths leave the Docker container behind with no record pointing at it (see #425). These paths affect the local Docker sandbox mode only.

In the CLI (go/cmd/amika/sandbox/sandbox_create.go), `store.Save` failing returned `sandbox created but failed to save state` with no cleanup: `rollbackVolumes()` exists in the same function and runs on every earlier failure, and `rb.Disarm()` sits just below the error return, but neither ran on this path. In the service (go/pkg/amika/service.go), `s.sandboxes.Save` failing returned `ErrInternal` while skipping even the setup-script and git-clone temp dir cleanups that run on every earlier failure.

The outcome is an orphan: retrying fails with a docker name collision, and no amika command can reach the leftover container or its `amika-git-*` / `amika-rwcopy-*` volume refs.

## Fix

- CLI `store.Save` failure now detaches volume refs, rolls back the rwcopy mounts via the existing `rollbackVolumes()`, and removes the container with `sandbox.RemoveDockerSandbox(name)` before returning. The original save error is kept in the message; if removing the container also fails, the error names the container so the user can remove it manually.
- Service `s.sandboxes.Save` failure now runs `cleanupSetupScript()` and `cleanupGitRepo()` and removes the container, wrapped as `ErrInternal` with the same explicit fallback message.

## Tests

New `TestSandboxCreateSaveFailureCleansUpContainer` in go/test/integration/cli: it occupies the sandbox state file's path with a directory so the record store can be neither read nor written, runs a bare create (from a non-git working directory so the sandbox record is the only state the create has left to save), asserts the create fails with a save-state error, and asserts `docker ps` shows no container under the requested name.

Verified against a real local Docker daemon (Docker 29.6.1):

- `AMIKA_RUN_DOCKER_INTEGRATION=1 go test ./test/integration/cli -run TestSandboxCreateSaveFailureCleansUpContainer` passes on this branch (2.2s) and fails on `main` with `container "amika-savefail-…" left behind after save failure`.
- Both create paths reproduce by hand: with the record save forced to fail, `main` leaves the container running via the CLI and via `POST /v1/sandboxes`; this branch removes it in both, with the CLI exiting 1 and the API returning 500.

The first version of this test made the whole state directory read-only. That also blocks the rwcopy staging that agent credentials use, so on a machine with agent credentials present the create failed while staging mounts, before the save, and the new assertion never ran. Occupying the state file path instead confines the fault to the sandbox record.

Fixes #425

### DIFF
--- a/go/cmd/amika/sandbox/sandbox_create.go
+++ b/go/cmd/amika/sandbox/sandbox_create.go
@@ -286,7 +286,15 @@ var sandboxCreateCmd = &cobra.Command{
 			Branch:      effectiveBranch,
 		}
 		if err := store.Save(info); err != nil {
-			return fmt.Errorf("sandbox created but failed to save state: %w", err)
+			rollbackVolumes()
+			// The sandbox record is the source of truth for every other
+			// command, so a saved-failed create must not leave the container
+			// (or its mounts) behind: the next run would fail on "docker
+			// name already exists" with no way to reach the orphan.
+			if rmErr := sandbox.RemoveDockerSandbox(name); rmErr != nil {
+				return fmt.Errorf("sandbox created but failed to save state: %v; cleanup of container %q failed: %w", err, name, rmErr)
+			}
+			return fmt.Errorf("sandbox created but failed to save state, the container was removed: %w", err)
 		}
 		rb.Disarm()
 
--- a/go/pkg/amika/service.go
+++ b/go/pkg/amika/service.go
@@ -181,7 +181,15 @@ func (s *serviceImpl) CreateSandbox(_ context.Context, req CreateSandboxRequest)
 		Branch:      req.Branch,
 	}
 	if err := s.sandboxes.Save(info); err != nil {
-		return Sandbox{}, fmt.Errorf("%w: %v", ErrInternal, err)
+		cleanupSetupScript()
+		cleanupGitRepo()
+		// Like the CLI path, never leave a container with no sandbox record:
+		// retries would collide on the container name with no way to reach
+		// the orphan.
+		if rmErr := sandbox.RemoveDockerSandbox(name); rmErr != nil {
+			return Sandbox{}, fmt.Errorf("%w: failed to save sandbox state: %v; cleanup of container %q failed: %v", ErrInternal, err, name, rmErr)
+		}
+		return Sandbox{}, fmt.Errorf("%w: failed to save sandbox state, created container removed: %v", ErrInternal, err)
 	}
 	publicMounts := toMounts(info.Mounts)
 	return Sandbox{
--- a/go/test/integration/cli/sandbox_create_save_failure_test.go
+++ b/go/test/integration/cli/sandbox_create_save_failure_test.go
@@ -1,0 +1,54 @@
+package cli_test
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+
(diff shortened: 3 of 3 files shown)


warun7:
Reopened. I closed this by mistake.

I also fixed the integration test, which I want to flag since it was the weakest part of the original PR. It made `AMIKA_STATE_DIRECTORY` read-only to force the save to fail, but the state directory is also where rwcopy staging for agent credentials happens. On any machine with agent credentials present the create failed while staging mounts, before reaching the save, so the new assertion never ran the test failed on both `main` and this branch under my normal environment. It only went green with an empty `HOME`.

The test now occupies the sandbox state file's path (`sandboxes.jsonl`) with a directory instead, so the record store can be neither read nor written while volume, rwcopy and file-mount state elsewhere in the state directory still works. That confines the fault to the save the test is about.

With real Docker 29.6.1:

- this branch: `AMIKA_RUN_DOCKER_INTEGRATION=1 go test ./test/integration/cli -run TestSandboxCreateSaveFailureCleansUpContainer` passes in 2.2s
- `main`: the same test fails with `container "amika-savefail-…" left behind after save failure`

I also reproduced both create paths by hand against local Docker mode: with the record save forced to fail, `main` leaves the container running via the CLI and via `POST /v1/sandboxes`; this branch removes it in both (CLI exits 1, API returns 500).

PR #428, opened 2026-09-07
Append the managed Include as overridable defaults

jdc123:
Stacked on #429 (review that first; this PR's diff is the second commit only).

Replaces the earlier version of this PR, which fought the user's config to make Amika's settings win. This does the opposite, per review feedback: Amika's block becomes **overridable defaults**, and we never reorder anything.

## What changed

The `Include` line went at the *top* of `~/.ssh/config`, which made the managed block win every option it named — ssh_config(5) resolves each option to the [first value it finds](https://man.openbsd.org/ssh_config). Wrong default for a file that governs every SSH connection the machine makes, not just Amika's. It now goes at the **end**, with a comment saying which side wins:

```
# Amika's defaults for sandbox hosts (*.amika). ssh uses the first value it
# finds for each option, so settings ABOVE this line override them; settings
# below it do not. "Host *" scopes the include so it is not swallowed by
# whatever Host block happens to precede it.
Host *
Include amika.conf
```

An Include that already exists is left exactly where it is. No reordering.

## The `Host *` guard is not decoration

This bit me while building it, so it's worth flagging. An ssh config block has no terminator — every directive belongs to the preceding `Host`/`Match` until the next one. So appending a bare `Include` to a config that ends inside `Host myserver` makes the include **conditional on connecting to myserver**, and the managed block silently never loads.

Measured with `ssh -G` on a sandbox alias, config ending in a `Host myserver` block:

| | bare `Include` appended | with `Host *` guard |
|---|---|---|
| `proxycommand` | *(absent)* | `amika plumbing ssh-stdio-proxy %h` |
| `identitiesonly` | `no` (default) | `yes` |
| `userknownhostsfile` | `~/.ssh/known_hosts` (default) | Amika's file |

The guard can't affect the user's other hosts, since the included file contains nothing but Amika's own `Host` blocks — verified that `myserver` still resolves to its own `user me` / `hostname example.com`.

## The trade-off, stated plainly

A wildcard stanza of the user's own now outranks Amika for sandboxes. Measured with `Host *` / `ForwardAgent yes` / `StrictHostKeyChecking no` above the include:

- `stricthostkeychecking false` — **their setting wins, so host-key pinning is off for sandbox connections**
- `proxycommand`, `user`, `identitiesonly`, `userknownhostsfile` — Amika's, since their stanza didn't set those

That's the deliberate cost of not overruling the user's config. Two notes on it:

- A `Host *` `ProxyCommand` in someone's config will now **break** `sandbox ssh` outright. Rare, but it will look like an Amika bug. The docs point at `ssh -G <alias>` as the diagnostic.
- The robust fix for anything we truly can't yield on is to pass it via `-o` on the ssh command line, which outranks every config file. Not done here; worth its own PR if we decide host-key pinning qualifies.

## What this PR no longer contains

The four client-side forwarding denials (`ForwardAgent no` etc.) are dropped, for two reasons:

1. **Already enforced server-side.** `amikad`'s loopback sshd — the daemon every `*.amika` connection terminates at — already sets `AllowAgentForwarding no`, `X11Forwarding no`, `PermitTunnel no`, `AllowTcpForwarding local` (`go/internal/amikad/sshd/manager.go`). The server refusing is authoritative; the client asking is irrelevant.
2. **Inert here anyway.** With the include last, either nothing else matches the alias (and OpenSSH's own defaults for those options are already `no`) or the user's stanza wins. Either way the lines do nothing.

So Codex's P1 on the previous version is moot: it was about the denials needing to outrank the user's config, and we're no longer trying to.

## Review follow-ups

Both live Codex findings were correct and are fixed:

- **`IdentityFile` accumulates, so "first value wins" was too broad.** Confirmed: with a wildcard `IdentityFile /user-key` above the include, `ssh -G` lists *both* it and Amika's key, while a scalar option (`ServerAliveInterval`) resolves to the user's value alone. The banner in the user's config, the code comment, and the docs now distinguish scalar from accumulating options, and the table no longer promises exactly one key.
- **A second-pass match block cannot override these defaults from either side.** Confirmed for both predicates: `Match final` and `Match canonical` (the latter whether or not `CanonicalizeHostname` is on) resolve to Amika's value above *or* below the include, since the reparse happens after the first pass assigned the scalars. Rather than enumerate predicates, the docs now state the rule: your override has to sit in a first-pass block, i.e. a plain `Host`/`Match` above the include, or `-o`. Left out of the in-file banner on purpose: more lines in the user's own config about directives almost nobody writes. Rendering the managed blocks with `Match final` would remove the inconsistency, but editors discover sandboxes by enumerating `Host` entries, so they have to stay `Host`.
- **An existing include is not repositioned, so the "appends, and your config wins" claim was unconditional and wrong for upgraded installs.** Anyone set up before this change has the include near the top, where Amika still wins. Documented as an explicit exception, with how to relocate it by hand.

Also removed a paragraph of mine that contradicted this page's existing forwarding contract (it said `-o ForwardAgent=yes` re-enables agent forwarding; the sandbox's sshd refuses it regardless). That contract is now stated once, with where it is enforced.

The two P1s on earlier revisions are moot: both were about the client-side denials needing to outrank the user's config, and this revision no longer tries to.

## Tests

`fmtcheck`/`vet`/`lint` clean; `go test ./...` passes except `internal/materialize`, which needs `rsync` (fails the same way on `main`).

New coverage: the include lands after existing content with its guard and banner; the guard survives a config ending inside a `Host` block; an existing include is left byte-for-byte where it is; the stanza states which side overrides.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -284,7 +284,8 @@ amika sandbox ssh -N -D 1080 my-sandbox
 ```
 
 Local (`-L`) and dynamic (`-D`) forwarding are supported. Remote forwarding
-(`-R`), agent forwarding (`-A`), and X11 forwarding are not.
+(`-R`), agent forwarding (`-A`), and X11 forwarding are not: the sandbox's own
+SSH daemon refuses them, so no local config or `-o` override enables them.
 
 `-o` after the subcommand is ssh's ssh_config option, so amika's
 `-o`/`--output` is not available there; written before `ssh` it is rejected
@@ -419,7 +420,7 @@ A copy naming no sandbox at all is simply forwarded to the system `scp`.
 `sandbox ssh`, `sandbox code`, and `scp` all connect by handing system OpenSSH
 a hostname such as `my-sandbox.sb_123.app-amika-dev.amika`. Every setting that
 connection needs comes from a block Amika generates in `~/.ssh/amika.conf`,
-which `~/.ssh/config` pulls in through an `Include` line at the top. The file
+which `~/.ssh/config` pulls in through an `Include` line. The file
 is regenerated from Amika's own state whenever a command prepares an SSH
 target, so edits to it are lost. `amika auth login` and
 `amika secret ssh-keygen` both write it; `auth login` names the files it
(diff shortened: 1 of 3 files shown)


docs/cli-reference.md
```diff
@@ -431,11 +440,31 @@ touched.
...
 | `StrictHostKeyChecking` / `UserKnownHostsFile`| `yes`, Amika's file  | Pin each sandbox host key in a dedicated file, so a change fails closed      |
 | `ProxyCommand`                                | `amika plumbing …`   | Carry the session over Amika's WebSocket transport instead of a TCP dial     |
+| `ForwardAgent` / `ForwardX11` / `ForwardX11Trusted` | `no`           | Keep your keys and your display out of the sandbox                           |
+| `PermitLocalCommand`                          | `no`                 | Stop a `LocalCommand` from running on your machine on connect                |
 | `ServerAliveInterval` / `ServerAliveCountMax` | `15`, `3`            | Notice a dead transport instead of hanging                                   |
 
-Anything the block does not set is answered by whatever else in your
-`~/.ssh/config` matches the hostname, since OpenSSH merges every matching
-block and keeps the first value it finds for each option.
+The last two rows turn things off, which needs explaining. OpenSSH resolves
```
dbmikus: nit: docs vomit from the AI

These are user-facing docs. Agent is too verbose here.

PR #451, opened 2026-09-14
Clamp E2B session timeout to the plan limit

jdc123:
## Problem

Creating an E2B sandbox fails at the provider with:

```
400: Timeout cannot be greater than 1 hours
```

E2B rejects a `timeoutMs` above the account's plan ceiling outright. `e2bTimeoutMs()` clamps the requested auto-stop interval to `E2B_MAX_TIMEOUT_MS`, but that constant was 24 hours — the paid-plan maximum — so it clamped nothing a caller could realistically ask for, and any interval over the plan limit reached E2B and 400'd.

The control plane always sends an interval (it falls back to a configured default rather than omitting the field), so this is not an edge case: with a deployed default of 300 minutes, every E2B create fails.

## Fix

Lower `E2B_MAX_TIMEOUT_MS` to one hour, the lowest ceiling a plan imposes, so an over-limit interval makes the sandbox stop sooner instead of failing the create. This mirrors what `VERCEL_MAX_TIMEOUT_MS` already does for Vercel, and the comment there says the same thing: clamp locally so `create` is never rejected for an over-limit timeout.

E2B pauses rather than deletes a sandbox on timeout (`onTimeout: { action: "pause" }`), so a clamped sandbox resumes on its next use — the clamp costs an earlier suspend, not any work.

## Test plan

- `vitest run src/providers/e2b/provider.test.ts` — extended the existing clamp test to pin the new ceiling and to cover a 300-minute request.
- Full package suite: 37 files / 383 tests pass.
- `tsc --noEmit`, `eslint`, `prettier --check` clean.


jdc123:
Closing: the 1-hour ceiling is our E2B account's plan limit, not a property of E2B (paid plans go to 24h), so hardcoding it into this shared package would be wrong for other consumers and would go stale if we change plans. Fixing it control-plane side in amika-mono instead, where the account/deployment config already lives.

